PR #4, opened 2017-04-20
Attempt at building a release .zip from grunt

klassenjs:
I am far from sure this is the best way to do things.  Submitted as a starting point.  

I would really prefer to exclude files rather than include them in the zip unless we only packaging dist - but then we need to copy more into dist such as the LICENSE and README and maybe demo and src (which start to feel odd).


klassenjs:
Tested build-release works on Windows too.

theduckylittle:
Totally understand. It's easy to get burned by something new being added and then being missed from the packager.  I'll merge this PR but maybe we can think about some sort of test script we can run to raise confidence ?

PR #15, opened 2017-04-25
Modal Dialog Components

theduckylittle:
1. The "Modal" component is a base class upon which other Modal dialogs can be based.
2.  Wrote an example "Alter" modal dialog as a sub-classed component.  
3. Added "API" hooks for dialogs in Application.
4. Wrote a test to prove the modal dialog works.
5. Fixed some versioning problems that prevented the test suite from properly running.

This branch will help get printing moving along.

### DIFF
--- a/demo/test.js
+++ b/demo/test.js
@@ -91,6 +91,4 @@ app.loadMapbook({url: 'mapbook.xml'}).then(function() {
     tracker.startTracking();
 
     showTab('catalog');
-
-    app.alert('This is an alert message!');
 });
--- a/demo/test.js
+++ b/demo/test.js
@@ -91,5 +91,6 @@ app.loadMapbook({url: 'mapbook.xml'}).then(function() {
     tracker.startTracking();
 
     showTab('catalog');
-   
+
+    app.alert('This is an alert message!');
 });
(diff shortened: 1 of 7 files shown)


klassenjs:
My biggest concern with Modal is it seems like it is a bit on the application side than the library side as it is taking a stand on UI elements.  At the same time, I know we need something that we can rely on that is available.  

Is there a way to override Modal to fit in with whatever toolkits a user might be using?  This might be particularly important for mobile/touch/smaller screen use cases.

theduckylittle:
I did take that into concern during the design phase.

I think if you look at the actual DOM tree that is created by the Modal that it's *very* simple to tweak. Mobile clients could adjust the modal to use the full screen.  And I tried to use a generic default style that would fit in with most of the major toolkits (Material, Bootstrap, jQquery Mobile).

klassenjs:
Maybe this should be split somehow between library and application.  Actually, I kind of wonder if print is the similar...  where the functionality needed to support printing (getting a hires canvas of the map contents made to a particular spec, getting legend info, maybe selected feature info, etc) should be in the library but leave it to the application to implement the UI and maybe templates.

theduckylittle:
@klassenjs approved this on a backchannel discussion.  Merging.

PR #20, opened 2017-05-02
Add PDF Printing

theduckylittle:
This uses jsPDF to add client-side based printing to GeoMoose.  There's a lot to unpack here but I'll try and hit the highlights:

- There are special print modules for creating a hidden background map for pulling in images of the required size for printing.
- The print action is user define-able which also means that things like layouts and other quirks can be changed by the user in the app-js.
- There is now a print reducer.  This is so that the printing image can be generated "asynchronously" and pipelines between different components (specifically the PrintImage and the PrintPreviewImage).
- Layouts actually now feature a JSON-esque meta language for adding elements to the print out: map, image, ellipse, rect, and text elements.  Along with colors, thicknesses, and positions.


klassenjs:
This branch appears broken (tested index.html in Windows/Chrome).
The catalog doesn't render until you click on the "Catalog" tab heading.  Console log says `Minified React Error #37` => Target Container is not a DOM element.

klassenjs:
Issue: div#print-preview missing in demo/index.html

PR #21, opened 2017-05-02
Renamed contributors to AUTHORS to fit in with npm.

theduckylittle:
Cool kids call it AUTHORS. We want to be cool.


klassenjs:
Verify the format is ok.  I'm not sure about markdown and/or comments. The docs say AUTHORS is a text file with a line per author of the format:
NAME <email> (URL)

The reference I have is vague near the end of https://docs.npmjs.com/files/package.json (contributors references the AUTHORS file.)  I am not sure if the case/+.md stuff applies to it the same as the README and friends mentioned elsewhere.

theduckylittle:
I've updated AUTHORS based on the NPM guidelines.  Instead of linking to the license, I embeded it in "comments".

Can we leave the README all ".md"-y like?  it makes it easier to read on the repo front-page and the syntax isn't so cluttered that it makes it hard to read in plain text as necessary.

PR #26, opened 2017-05-02
Use relative paths in demo/test.html

klassenjs:
So it can work outside the webpack-server/docker environment.


klassenjs:
Taking a step back: What purpose does test.html (still) serve?

Rational: My understanding test.html is meant to be used with webpack serve auto-rebuilds.  But webpack auto-rebuilds ./dist/ when there is any interesting change (so all the stuff should be in ./dist anyway).  And it is no longer the case that test.html can run without building (it depends on ../dist/geomoose.js).  At most, I'd think we'd need to make sure ./src/services are in webpack's watch list.

However, separate files (index.html and test.html) are a maintenance burden and test.html could easily mask an error where not all of the build components are put in place.

klassenjs:
To answer my question: test.html refrrences un-minified versions of the resources so debuggers work.

PR #28, opened 2017-05-04
Created a modal for uploading files.

theduckylittle:
The old tool threw up a file open dialog but didn't give
the user any context as to what was happening. With the new modal dialog
componenet it made sense to try and help the user a bit more.

Changes:
1. Moved the upload tool out of tools.jsx and into its
   own module.  This cleans up the organization as there is
   a dependent Modal class now in upload.jsx.
2. Small tweak to the modal CSS that fixes an embedding issue.
   Now it's possible to put the dialog div nearly anywhere in the DOM
   and it will render correctly.


klassenjs:
Not sure this needs to be done within this pull request, but I found some gotchas while trying it out.
* The modal doesn't say what types of files are understood.
* Unsupported files are uploaded and ignored (no indication of success or failure).
* Bonus points: can the open dialog only allow supported extensions?
* It'd be cool if we could identify on the uploaded features.

theduckylittle:
Re: 1. I'm open to some suggested text but at least letting the user know what's going on was a substantive improvement.
Re: 2 + 3: This is possible but would be more work and filing a fresh issue would be helpful.
Re 4: We can but it requires configuration out of scope of the dialog box.

Issue #9, opened 2017-04-20
Add a Length Input Type


theduckylittle:
GeoMOOSE 2.X had a special type of service input that would allow a user to specify a distance in their preferred units.  gm3 needs the same capability.

Acceptance criteria:
1. A new input type that displays both a text field for the distance and a drop down for units.
2. The new class should be placed in src/gm3/components/selectInputs/length.jsx and appropriately extend TextSelect.
3. gm3/util.js should get a distance conversion function that takes in a length, a source units and a destination units.  There are already functions that convert to/from meters.  Those functions can stay there for now but we should have a new one that follows this prototype: `export function convertLength(length, srcUnits, destUnits)` where srcUnits/destUnits are one of `m,in,ft,mi,km,yd`.  

This is the old 2.X implementation:
 - The input: https://github.com/geomoose/geomoose-js/blob/master/geomoose/GeoMOOSE/ServiceInputTypes.js#L598
- The conversion function: https://github.com/geomoose/geomoose-js/blob/master/geomoose/geomoose.js#L808

elil:
Chains might be useful too.  In 2.#, I think we're using [OpenLayers](https://github.com/openlayers/ol2/blob/f4a4090814f240414548e8eda55070d16f3d3901/lib/OpenLayers/Util.js#L1057).

theduckylittle:
In 2.x we were using the OpenLayer's lookup tables.  If someone has time to submit an addition for chains then it's cool with me.

elil:
:thumbsup: Thanks!

theduckylittle:
I've merged in @klassenjs 's works but that *does not* complete the ticket.  The Input-length type will still be needed. This does give a great example on the proper units conversion. Thanks again!

PR #30, opened 2017-05-04
Length Service Input

CaitW:
Adds a Length service input type per issue #9 

* Does the select input need an ID field? What do we use the ID field for?

### DIFF
--- a/src/gm3/components/serviceInputs/length.jsx
+++ b/src/gm3/components/serviceInputs/length.jsx
@@ -1,0 +1,78 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2016-2017 Dan "Ducky" Little
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+import React from 'react';
+import TextInput from './text.jsx';
+import { convertLength } from '../../util.js';
+export default class LengthInput extends TextInput {
+    constructor(props) {
+        super(props);
+        this.units = [{
+            label: 'Feet',
+            value: 'ft'
+        }, {
+            label: 'Yards',
+            value: 'yd'
+        }, {
+            label: 'Miles',
+            value: 'mi'
+        }, {
+            label: 'Inches',
+            value: 'in'
+        }, {
+            label: 'Meters',
+            value: 'm'
+        }, {
+            label: 'Kilometers',
+            value: 'km'
+        }, {
+            label: 'Chains',
+            value: 'ch'
+        }];
+        this.state = {
+            value: props.field.default ? props.field.default : 0
+        };
+    }
+    onChange() {
+        let lengthInput = parseFloat(this.lengthInput.value);
+        let unitInput = this.unitInput.value;
+        let lengthInMeters = convertLength(lengthInput, unitInput, 'm');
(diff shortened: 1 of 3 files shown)


src/gm3/util.js
```diff
@@ -436,3 +436,21 @@ export function metersAreaToUnits(meters, units) {
...
+        'ft': 0.3048,
+        'yd': 0.9144,
+        'mi': 1609.347,
+        'in': 0.0254,
+        'm': 1,
+        'km': 1000,
+        "ch": 20.11684
+    };
+    let convertedLength = length * equivalentMeters[srcUnits] / equivalentMeters[destUnits];
+    if (typeof convertedLength === "number" && convertedLength >= 0) {
```
theduckylittle: This check should be removed.  Negative distances are a real thing and returning NaN is also valid and should be left up to the calling function to determine how to best handle that.

theduckylittle:
@CaitW Please checkout my comment, everything else looks good,.

klassenjs:
Why are we duplicating the conversion table in convertLength and
metersLengthToUnits?

Is the hash lookup or the switch faster?

theduckylittle:
That's a cleanup step we need to make.  I did the previous "metersLengthToUnits" function as a quicky  for making the measure tool work properly.  Ostensibly, metersLengthToUnits should be retired in favor of the new function.  But that's out of scope of the issue.  I can write up an issue or just mentally note that after this merge is done the measure tool should be converted.  

Also the switch vs hash is an interesting question.  To my understanding the switch is probably faster and more memory efficient but the code in convertUnits is fairly elegant.  Putting the hash in a more permanent memory area (as a const ahead of the function) would probably increase the speed as there wouldn't be the need to reinitialize the object with every function call.  (Lazy dev coming... )Of course, the JIT/VM may just do that automatically and not really be an issue. (end)

klassenjs:
Then I would recommend the documentation of the constants be included in
the new function.

On May 5, 2017 08:46, "theduckylittle" <notifications@github.com> wrote:

That's a cleanup step we need to make. I did the previous
"metersLengthToUnits" function as a quicky for making the measure tool work
properly. Ostensibly, metersLengthToUnits should be retired in favor of the
new function. But that's out of scope of the issue. I can write up an issue
or just mentally note that after this merge is done the measure tool should
be converted.

Also the switch vs hash is an interesting question. To my understanding the
switch is probably faster and more memory efficient but the code in
convertUnits is fairly elegant. Putting the hash in a more permanent memory
area (as a const ahead of the function) would probably increase the speed
as there wouldn't be the need to reinitialize the object with every
function call. (Lazy dev coming... )Of course, the JIT/VM may just do that
automatically and not really be an issue. (end)

—
You are receiving this because you commented.

Reply to this email directly, view it on GitHub
<https://github.com/geomoose/gm3/pull/30#issuecomment-299468803>, or mute
the thread

theduckylittle:
re-reading the PR code post-coffee -- there really should be more comments than just the license header.

Please ensure that each class has a description of what it does and each function which is non-standard (IE, not render, onChange, etc.) is given a proper description, parameter values, and expect return value description.

src/gm3/components/serviceInputs/length.jsx
```diff
@@ -0,0 +1,94 @@
...
+
+/**
+ * Service input type for specifying length
+ * - Renders a numeric text input and a select-list of units
+ * - User-specified length is converted to its equivalent in meters and stored
+ *   in the component state
+ *
+ * @extends TextInput
+ */
+import React from 'react';
```
theduckylittle: New line

src/gm3/components/serviceInputs/length.jsx
```diff
@@ -0,0 +1,94 @@
...
+import TextInput from './text.jsx';
+import { convertLength } from '../../util.js';
+export default class LengthInput extends TextInput {
```
theduckylittle: new line

src/gm3/components/serviceInputs/length.jsx
```diff
@@ -0,0 +1,94 @@
...
+            value: 'km'
+        }, {
+            label: 'Chains',
+            value: 'ch'
+        }];
+        this.state = {
+            value: props.field.default ? props.field.default : 0
+        };
+    }
+    onChange() {
```
theduckylittle: new line

src/gm3/components/serviceInputs/length.jsx
```diff
@@ -0,0 +1,94 @@
...
+        let lengthInput = parseFloat(this.lengthInput.value);
+        let unitInput = this.unitInput.value;
+        let lengthInMeters = convertLength(lengthInput, unitInput, 'm');
+        this.setState({ value: lengthInMeters });
+        this.setValue(this.getName(), lengthInMeters);
+    }
+    /**
```
theduckylittle: new line

src/gm3/util.js
```diff
@@ -436,3 +436,23 @@ export function metersAreaToUnits(meters, units) {
             return meters;
     }
 }
+/** Converts numeric lengths between given units
```
theduckylittle: new line

src/gm3/util.js
```diff
@@ -436,3 +436,23 @@ export function metersAreaToUnits(meters, units) {
...
+ *
+ * @param {number} length - Length
+ * @param {string} srcUnits - Source unit
+ * @param {string} destUnits - Destination unit
+ * @return {number} Converted length
+ */
+export function convertLength(length, srcUnits, destUnits) {
+    // US survey feet, miles
+    const equivalentMeters = {
```
theduckylittle: As a small optimization (and to make the constants available to others, this should get moved OUT of the function and put into it's own declaration.  E.G. "const EQUIVALENT_METERS = { ..}" on line 439.

theduckylittle:
"Don't fear the void".  I added comments suggesting newline to help break up the visual flow the code a little bit more.  Not having the blank lines after function declarations can make the code look denser than it really needs to be and harder to visually breakup by others.

PR #33, opened 2017-05-09
Add Column Filters to the Grid

theduckylittle:
It is now possible to add a filter to results (instead of just deleting them). The user can now configure the "filter" property of a "column" when defining the grid's header row.  The three types of filters that are current supported are:
1. `list` - Settings `values` lets the administrator set an enumerated list of values, leaving it undefined will have the element select all of the values from the list.  Lists are presented as checkboxes.
2. `range` - Users are offered a "min" and a "max".  The user can specify only max, only min, both or neither.
3. `equals` - This is a simple "does this column = whatever is in the textbox" type of filter.


theduckylittle:
Screen shot for posterity:
image

klassenjs:
I'm not entirely clear on how Clear and Set work.

1)Select Features, pick two (doesn't really matter, two is simple).
2) Click the Filter icon by PIN
3) Uncheck one item, click Set -> Unchecked item disappears from map and table.
4) Click Filter icon by PIN
5) Uncheck other tiem, click Set -> Both items disappear
6) Click Filter icon by PIN
7) Check both items, click Set -> Both items reappear
8) Click Filter icon by PIN
9) Click Clear -> Dialog closes, both items remain (wasn't expecting box to close, was expecting more of a select all/none type behavior)
10) Click Filter icon
11) Both items now unchecked (odd since both items are on map and in table)
12) Select one item, Click Set -> both items on map, no items in table (odd)
13) Click Filter icon
14) Same as 11
15) The only way back to normal that I've found is deleting the select from the super tab and making a new one.

theduckylittle:
'Clear' is meant to unset the filter no matter what's selected -- it removes all settings for that filter, regardless of what's in the text boxes or checkboxes.

This was really a bug with how clear was handling the values when you left the dialog. It should've restored all the checkboxes.

Issue #25, opened 2017-05-02
Tests don't run on Windows


klassenjs:
`npm test` fails to run any tests.

```
C:\ms4w\apps\gm3>npm test

> geomoose@3.0.0-beta test C:\ms4w\apps\gm3
> node node_modules/jest/bin/jest.js --coverage ./tests/

No tests found
  70 files checked.
  roots: C:\ms4w\apps\gm3 - 70 matches
  testMatch: **/__tests__/**/*.js?(x),**/?(*.)(spec|test).js?(x) - 4 matches
  testPathIgnorePatterns: \\node_modules\\ - 70 matches
  testPathPattern: "./tests/" - 0 matches
----------|----------|----------|----------|----------|----------------|
File      |  % Stmts | % Branch |  % Funcs |  % Lines |Uncovered Lines |
----------|----------|----------|----------|----------|----------------|
All files |  Unknown |  Unknown |  Unknown |  Unknown |                |
----------|----------|----------|----------|----------|----------------|
```

theduckylittle:
Here's a PR: https://github.com/geomoose/gm3/pull/36

theduckylittle:
Closing after the PR merge.  Tests on windows are working!

PR #36, opened 2017-05-10
Fixed path issue with tests on windows.

theduckylittle:
`./tests/` wasn't a valid path on windows.  Dropping the relative paths fixes that.


theduckylittle:
Refs: #25

PR #41, opened 2017-05-12
OSM Geocoder

CaitW:
Effectively the same as the Bing Geocoder, updated for the OSM API.

### DIFF
--- a/src/services/geocode-osm.js
+++ b/src/services/geocode-osm.js
@@ -1,0 +1,120 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2016-2017 Dan "Ducky" Little
+ *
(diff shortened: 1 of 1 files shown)


theduckylittle:
I'm getting an error on line 99... says "feature" is undefined.   Happens when I do the "second" search of an address.

src/services/geocode-osm.js
```diff
@@ -0,0 +1,120 @@
...
+                });
+
+                let features = [];
+                for(var i = 0, ii = results.length; i < ii; i++) {
+                    var r = results[i];
+                    features.push({
+                        type: 'Feature',
+                        geometry: {
+                            type: 'Point',
+                            coordinates: [feature.lon, feature.lat]
```
theduckylittle: Feature doesn't appear to be defined anywhere?

PR #42, opened 2017-05-13
Set the OSM geocoder to be the default service.

theduckylittle:
Accounts for @CaitW's changes and puts the OSM geocoder as the default geocoder.


theduckylittle:
Sneaking. This. Through.

PR #1, opened 2017-04-16
Add Travis CI Configuration File

theduckylittle:
Adds a travis configuration that will check the code lint and run the basic tests.

PR #43, opened 2017-05-15
Hash Track!

theduckylittle:
The weird, the wonderful, the hash track!

This is the first cut of the hash track feature. It converts the current state to a "#" encoded URL.  On reload, it restores that state.  Changes needed to the app:

1. This changes how the default view is set.  Application now has a function to handle that instead of relying on the Map component.
2. The addition of the hash tracker class itself and properly indexing it.

Some difference over the implementation in 2.X:
1. The URL changes are "one way". If you change the hash it will not route.  The only time the URL is parsed is on startup.  It sounds problematic but I don't think it's a huge loss to the user.  Plus it prevents having to do lots of circular logic checking.
2. Right now it is possible to turn highlight "off" in the on= list and not turn it back on.  After the merge, this should probably get filed as a bug.  But I think having layers that are "permanently on" would merit discussion before implementation.
3. "loc=" is storing z,x,y.  This is not alphabetically correct but more inline with other conventions. I'm not  sold on this but it seems trendy.


klassenjs:
This is going to sound weird, but I can reproduce it on feature/hash-track and switching between master and feature/hash-track (and rebuilding) makes it go away/come back...  I am noticing high DISK usage on mouse movement over the map with feature/hash-track (many seconds worth, in Firefox).  (CPU usage either way is ~ 40%, Memory usage is similar ~5%).  The process doing IO is `firefox-esr [mozStorage #1]`.

klassenjs:
Also not really able to test this well due to the above slowness, but what happens if the hash data is invalid?

theduckylittle:
That's really weird, I'm not seeing any impact on Chrome.

I can add some parsing checks but it is possible to put the application in a "bad state" if you change the hash and hit reload.

klassenjs:
One more to test: does this make the browser forward/back buttons function?

klassenjs:
The slowness happens immediately on loading the feature/hash-track branch.  No funny business on my part yet.  Moving the mouse in a circle over the map for 5 seconds can create a minute or so of disk activity (on 7200RPM drive, ext4, 5MB/s, so it must be a huge seek load).

theduckylittle:
It *does not honour* changes to the hash (which the forward-back buttons trigger) See "Point #1" in the first post.  If this is desired as a follow up, we'll do it but by ignoring it, there's a lot of logic that is simplified.

theduckylittle:
Found the offending bits.  Trackers are actually subscribed to the entire state-tree.  This is because they could, ostensibly, always be tracking state.  The mouse position from the map is also tracked in state.  Mostly so that it can be used to render the current position in other components.  This led to a bit of a problem: everything the mouse moved the tracker was being called and it was triggering a false event.

The window hash would get updated because "this.lastHash" was always set to ''.  This fixes that so the check actually works and memory consumption goes down considerably.

theduckylittle:
FWIW, my system must be a bit more powerful.  I was able to get FF to hit 120% CPU usage by just moving the mouse around but it never *stopped* or froze wholesale.  So @klassenjs  this should get retested by you.

klassenjs:
Again, it was disk bound not cpu or memory bound.  Given the description my
guess is Firefox was writing the location bar history to disk every time it
changed.

klassenjs:
Will retest when I get back.

klassenjs:
Latest commit seems to have fixed it.

BTW: Chromium did it too, but not as bad.  It does a lot of writes on it's DB Thread initially on movement but then stops.

klassenjs:
I really think it should honor the forward and back buttons as it is adding the state that activates them.  Maybe something that can wait for a future PR.  I understand the event loop complexities though.

theduckylittle:
And now with back/forward support.

klassenjs:
Almost there.   Catalog and Visible tabs aren't listening.  Ex. Turn on a layer and click back.  Map layer turns off, catalog & visible don't uncheck.  Tested FF and Chrome.

theduckylittle:
@klassenjs  - This has the fixes for the tabs.  It's a complete and proper fix that has the catalog now listening to the mapSources section of the state tree instead of carrying it's own extra state.

klassenjs:
If the layer is turned off by going back, you can't turn it on with the checkbox.  If you go forward again so it turns on then the checkbox works normally.

theduckylittle:
Fixed. I still has a CATALOG.LAYER_VIS action being issues instead of just directly issuing map-source actions.

klassenjs:
Cool.  Seems I have run out of ways to break it.

PR #58, opened 2017-05-16
Download tool!

theduckylittle:
Layers can now be downloaded as GeoJSON or KML when
configured for the download tool.


klassenjs:
Looks good to me.  Merging.

PR #62, opened 2017-05-18
Fix for broken WMS legend graphics.

theduckylittle:
Forgot to set the default FORMAT parameter. That was causing the legend to
not generate.  Defaults to PNG, this can be overriden in the map source
definition.

### DIFF
--- a/src/gm3/components/layers/wms.js
+++ b/src/gm3/components/layers/wms.js
@@ -101,6 +101,7 @@ export function getLegend(mapSource, mapView, layerName) {
         'VERSION': '1.1.0',
         'WIDTH': '250',
         'LAYER': layerName,
+        'FORMAT': 'image/png',
     }, mapSource.params);
 
     const images = [];


klassenjs:
Is ending a hash with a ',' legal?

===

```
     const params = Object.assign({
         'REQUEST': 'GetLegendGraphic',
         // 'SCALE': TODO: get a scale hint from the map.
         'SERVICE': 'WMS',
         // TODO: Does this need to be passed in? Check by server-type?
         'VERSION': '1.1.0',
         'WIDTH': '250',
         'LAYER': layerName,
         'FORMAT': 'image/png',
     }, mapSource.params);
```

The SERVICE=WMS parameter is part of the spec.

SCALE hint is needed at least for MapServer scale dependent layers (legends can change based on scale, we spent a long time getting this right in 2.x, I don't want to backtrack on it now.)

FORMAT PNG is a reasonable default, but should be settable in the map-source (WMS spec says what is offered is advertized in the GetCapabilities and image/png might not be on the list).

Pinning version at 1.1.0 seems odd but in any case should match the GetMap request.

The above probably should be made into a bug issue.

theduckylittle:
1. The "," is legal with the use of babel. IE doesn't support it but the compiler cleans it up.
2. So back to SCALE, we don't have a good way of passing it in.  It's not calculated well.  My understanding is that without the SCALE hint it should just return ALL of the legends.  This isn't really a step backwards.  Also, the lack of SCALE shouldn't hold back this PR which fixes a clearly broken problem.
3. If FORMAT is set in the map-source it will override the PNG setting.  which is why mapSource.params is mixed in AFTERWARDS.

klassenjs:
1) Weird, ok.
2) This will be a problem, but agree it shouldn't hold up the merge (I tried to indicate that but see how I was unclear).
3) Nice.

Issue #49, opened 2017-05-16
Layer Re-ordering Controls


theduckylittle:
1. We need to add "Up" and "Down" controls to the catalog / Visible Layers to allow users to visually sort the layers.
2. Visible Layers should be ordered by the layer's Z-order.

PR #65, opened 2017-05-18
Feature/layers up and down

theduckylittle:
https://github.com/geomoose/gm3/issues/49

1. Visible layers is now sorted by z-order.
2. The map now listens to z-order changes.
3. There are now up/down tools available for layers in the catalog (and are forced on in the visible layers tab).


klassenjs:
Layer order not preserved by HashTrack

klassenjs:
Otherwise works as advertised.

klassenjs:
It was decided that the z-order tracking is a HashTrack enhancement. #66

Issue #45, opened 2017-05-16
Fade and Unfade Layers


theduckylittle:
Implement opacity controls for individual layers.

PR #67, opened 2017-05-22
Added opacity controls!

theduckylittle:
New FadeTool and UnfadeTool along with all the necessary
supporting changes.

Supports: https://github.com/geomoose/gm3/issues/45


klassenjs:
Code looks good and test good.  Do you want to add it to the example mapbook before merging?

theduckylittle:
I didn't know what a good example would be? I used the AGS layer for testing.  Obviously, adding it as an example would be a good idea but I didn't really know which layer to "pick".

theduckylittle:
Maybe just throw it on the sketch layer since it has the full spread anyway?

klassenjs:
It doesn't have to be one layer.  Personally my choice would be add it to all the backgrounds to fade between them.

theduckylittle:
Added examples to the set of background layers.

Issue #60, opened 2017-05-16
"Ending" Map Interaction


theduckylittle:
There are cases where the user will switch away from the Super tab but leave an interaction with the map turned on that is not simply "navigate." 

There should be a big button on the map that says, "Stop [ Interaction ]" that cancels out any interaction the user may be performing at any time.  This gives a clear indication what tools is selected on the map even if the Super tab is not open at the time.

I think this could be a big usability win and really provide clarity to what tool is active when.

PR #70, opened 2017-05-23
Signal a definite way to end interacting with the map

theduckylittle:
This is addressing ticket #60 .  It presents the user with a way to stop whatever tool is open.

Definitely some room for improvement.  If there's no major functionality objections I'd like to see this merged so that we can start the feedback cycle on it. (Specifically, I'd like @elil 's eyes on this one.)

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,6 +61,7 @@
     "gruntify-eslint": "~3.1.0",
     "jest": "^19.0.2",
     "jspdf": "^1.3.3",
+    "jsts": "^1.4.0",
     "less": "^2.7.1",
     "mapskin": "^1.0.0",
     "markup-js": "^1.5.21",
--- a/src/gm3/components/map.jsx
+++ b/src/gm3/components/map.jsx
@@ -657,14 +657,14 @@ class Map extends Component {
             Point: 'End Drawing',
             Select: 'End Selection',
         }[type];
-       
+
         // helpful default behaviour for when the description
         //  is not defined.
         if(typeof(tool_desc) === 'undefined') {
             tool_desk = 'End ' + type;
         }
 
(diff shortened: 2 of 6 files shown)


klassenjs:
Draw tools aren't working for me with this PR.  I am seeing `i.append is not a function` in the console.

klassenjs:
Looks like the error is in createStopTool at elem.append(button);

elil:
This looks great.  It seems mostly oriented towards "drawing" (or "Selection of existing drawings").  It seems to only apply to the drawing phase of tools that have drawing phases (Measure, Identify, Select Features).  When these tools are not in a drawing phase (like when none of the drawing tools are engaged), there is is no "End" option.  

This emphasis on drawing is probably why it says, "End Drawing" (or "End Selection").  Although this is technical correct and very articulate, "Cancel" might be a more universally understood message.

Ending interaction (or "Cancel") may be expandable to other non-drawing or non-selection of existing drawings tools (Search, non-drawing phases of Measure, Identify, Select Features, etc).  

See also #81, for general tab interaction and how this may tangentially relate to this.

theduckylittle:
@klassenjs Do you have any objections to changing to the constant "Cancel" button instead of more detail? I'm okay with "Cancel" instead of the smart text.  In a non-pushed version of the code I was actually putting the draw tool name in the "stop" button but quit that.  "Cancel" does seem nice and consistent.

klassenjs:
I don't care what the button is labeled on the map.  
But it seems to me the button on the map should be related to things that change the map from the normal interaction and not necessarily a general cancel for anything.  I am unclear on the discussion what the actual scope is being discussed.  (Also IMHO it would be nice if escape key canceled drawing).

PR #71, opened 2017-05-23
Added "stop" control for drawings.

theduckylittle:
When drawing or selecting, the user can now "stop" or "end" the
drawing/selection process.  They can start it up again by clicking
the appropriate radio button from the service or a tool from a layer.


theduckylittle:
Refs: #60 , this is an updated PR which has the missing "Child" fix.

PR #74, opened 2017-05-24
Specify coordinate projection

CaitW:
This is still a bit of a work in progress. Questions/comments:

1. I didn't add a coordinate transform property (yet) in favor of proj/wkt strings

```Javascript
{
    label: 'UTM15',
    ref: 'EPSG:26915',
    // pretty
    projDef: '+proj=utm +zone=15 +ellps=GRS80 +datum=NAD83 +units=m +no_defs'
}
```

> We may need to add a function to application or util to "bridge the gap" to allow users to add a projection.

2. @theduckylittle would it make sense to keep projection definitions out of the coordinate display options? And then simply reference the defined projection in the coordinate display option:

```Javascript
// app.js

// define projection
app.addProjection({
    ref: 'EPSG:26915',
    def: '+proj=utm +zone=15 +ellps=GRS80 +datum=NAD83 +units=m +no_defs'
});

// then refer to it 
app.add(gm3.components.CoordinateDisplay, 'coordinate-display', {
    projections:  [
        {
            label: 'UTM15',
            ref: 'EPSG:26915'
        }
    ]
});
```

3. I added a `precision` property so users can truncate long coordinates, i.e.

```Javascript
{
    label: 'Lat,Lon',
    ref: 'EPSG:4326',
    precision: 3
}
```

### DIFF
--- a/examples/desktop/app.js
+++ b/examples/desktop/app.js
@@ -1,4 +1,4 @@
-/** Demo test application.  
+/** Demo test application.
  *
  *  WARNING! ACHTUNG! THIS IS FOR DEVELOPMENT PURPOSES ONLY!!!
  *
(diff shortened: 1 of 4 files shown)


klassenjs:
If we go with the (2) option, Proj4JS already has a way to add coordinate systems.  I would suspect deferring to that (or making addProjection a wrapper to that) would be better than reinventing it ourselves.  That way the same custom projection def can be used elsewhere in the application in the usual way.

CaitW:
@klassenjs I included a util function that wraps the proj4js library

theduckylittle:
This all works for me but can you restore the default projections (xy, usng, 4326).  I think having 4326 on there will serve as a reasonable example for someone who wants to change the default projections.

PR #75, opened 2017-05-26
Ol v4

theduckylittle:
Includes:
1. Removing all references to "ol." and is replaced by proper
   imports.

2. package.json has been updated to reflect the new version,
   and the new version of mapbox-to-ol-style.

3. getStyleFunction now refers to mapbox-to-ol-style.

4. Webpack config updated to include/exclude the appropriate
    libraries.

5. CSS now references "ol" instead of "openlayers".

6. Tests have been updated to work with the ES2015 style modules where applicable.


theduckylittle:
Backroom chatter on this ticket:
1. It does make the builds a lot bigger.
2. The builds take a lot longer.
3. We needed to fix windows because path matching wasn't working (node was not normalizing the paths to unix-style "/"es).

klassenjs:
The end result that the end users see is slightly smaller...
```
Old:
   748k geomoose.min.js
   496k ol.js
   ---------
   1244k total
New:
   1220k geomoose.min.js
    0k ol.js (doesn't exist anymore)
```
But it does take more time (85 vs 42 sec) and memory to (~980MB up from ~680MB) build as OpenLayers is now part of the build instead of tacked on the side after the build.

That said, except for babel/webpack continuing to use what is IMHO an insane amount of memory, I think being up to date with OpenLayers, and having more of the project in es2015 is a good thing.  This update is also something that is good to get done before we released 3.0 because there were some minor, but breaking changes to the examples, which implies we would have had to release a GeoMoose 4.0 very quickly.

klassenjs:
I have tested the build and application on Linux and Windows and I am +1 to merge but, as this is such a large changeset, I am going to hold off on clicking the big green button.

theduckylittle:
I've been playing with this for a while today and feel like we can do the merge.  Further, the more eyes on this the more we use it the more bugs we'll be able to identify and more quickly.

PR #76, opened 2017-05-26
Create a "smart" layer for rendering query results.

theduckylittle:
This PR accomplishes a few things:

1. Removes the custom filtering that I wrote to handle the grid filtering in favor of using the more feature complete mapbox "createFunction".  All the necessary components have been updated to reflect that.
2. Adds a new "transform" tag to layers so that a data type can be specified for a particular attribute.  When we get to feature editing this will probably need to come up again as we'll actually want more metadata about columns than this functionality provides.
3. The "highlight" layer has been removed in favor of an internal "results" source and layer.  That "results" layer is the one that is tied to the same display as the grid.  Ergo, both the grid and the layer will match when a filter is set for a query.
4. Multiple "layers" can now be defined for a vector map-source. The first "layer" is still used to set the type name for WFS requests, however, subsequent layers can define filters and styles for the layer. This also moves <style> into <layer> instead of being a part of <map-source>.
5. A small change to z-ordering was made.  A lot of the early code still assumes defaults will be set by the mapbook (this could use additional follow up work, actually) including the z-ordering.   There is a change that moves automatic z-ordering assignment out of the mapbook and into the "add" action.
6. "highlighting" now has some meta functions in application so that we can change the underlaying implementation more fluidly into the future.

### DIFF
--- a/examples/desktop/mapbook.xml
+++ b/examples/desktop/mapbook.xml
@@ -195,6 +195,8 @@
                 <td style="text-align: right">${{ properties.EMV_TOTAL | number | localize }}</td>
             </tr>
             ]]></template>
+
+            <transform attribute="EMV_TOTAL" function="number"/>
         </layer>
         <file>./demo/parcels/parcels.map</file>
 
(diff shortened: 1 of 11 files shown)


theduckylittle:
Jim says "Seems to work as advertised, I'm not confident about having it full tested."

Duck says, "I have 85% confidence. The other 15% will come from use."

Issue #82, opened 2017-05-27
Select Features by "Draw Multi-Point" does not draw or select


elil:
Trying to draw and use a Multi-Point for Select Features does not work.  It does not draw any points, it does not activate the "End Drawing" option, and it does not select anything.

theduckylittle:
Found the bug, I had a typo while handling the multipoint handler.  When #88  is merged this should work itself out.

elil:
This does now work as expected for multipoint select (and thus could probably be closed).  As noted in #88, there are still issues with multipoint buffer.

Issue #48, opened 2017-05-16
Buffered Selection


theduckylittle:
User's will want to be able to select features using a drawn geometry and a buffer.  They also want to select features from a layer, buffer those features, and select a set of features.  

The classic example: Buffer a parcel by 1,000 feet.

To do this, we're looking to use JSTS.  This is a big library and adds a lot to the deliverable size.  Care should be given to prevent the library from bloating much more.  Making JSTS optional may need to be on the table for folks who won't need the buffering functionality.

elil:
I've never understood the draw geometry and buffer option.  What is the purpose?  Why not just draw it bigger to begin with and not buffer it?  (buffering a line sort of makes sense).  

I understand the purpose of the select a geometry (or set of) from a layer and buffer it.

klassenjs:
Usually because you need to notify people within a certain distance of a
project area.

It might be more ideal to upload that geometry or select it from another
layer, but people don't always have it ahead of time.

On May 16, 2017 14:59, "Eli L Adam" <notifications@github.com> wrote:

I've never understood the draw geometry and buffer option. What is the
purpose? Why not just draw it bigger to begin with and not buffer it?
(buffering a line sort of makes sense).

I understand the purpose of the select a geometry (or set of) from a layer
and buffer it.

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/geomoose/gm3/issues/48#issuecomment-301897842>, or mute
the thread

theduckylittle:
PR merged, follow up tickets filed, this ticket can party on in closed statuis.

theduckylittle:
Option 2 can happen when you choose to buffer the results.

PR #88, opened 2017-05-27
Add client side buffering!

theduckylittle:
Change summary:

1. Adds JSTS dependency.  Unlike previous versions this exercises the new ES6 modules available for JSTS.  So while similar to the OL4 upgrade this does add to the build time it is now only pulling from JSTS what it really needs.  A more ambitious project (much later date) would be to pull the geometry models out of JSTS and create ones based on the openlayers geometries.  For now, we are using GeoJSON as a pass-through transport between JSTS/OpenLayers/GeoMoose.

2. I placed the buffering functionality into a dedicated JSTS module.  If someone really never needed the buffer capability then they could actually fork GeoMoose with a removed gm3.jsts and scrub the necessary functions from the map component and they could move on without that dependency.  Naming the gm3 module after the dependency breaks the tradition of "abstracting" functionality but as this could be a very large dep I wanted to make sure all of it's code was fairly isolated and obvious.

3. There is a new set of actions for handling buffers.  The buffer is actually set in the Map's state.  The map checks it for a non-zero value and then buffers the shape appropriately.

4. A "selection/selection" layer was added similar to the results layer.  This is handled internally as a display layer for the map's selection features.  It functionally works as a "preview" layer for the user's selection.  Word of warning: both the results/* layers and the selection/* layers do not have obvious user controls.  So when using HashTrack you can get the application into a bit of an odd state where the highlight/preview layers are off and they cannot be turned on.  I plan to file a follow up issue as to how that could be resolved.  It probably merits discussion with @klassenjs .

5. While "multipoint" is no longer broken (See issue #82 ) it has some odd behaviours that may not be expected when applying buffers.  I believe that JSTS is handling MultiPoint as a degenerate case and is only buffering the first point. 

6. Services are now configured with a "bufferAvailable" attribute.  When set to true, the buffer functionality will appear.

I think I've written almost as many lines as there is code in the PR. While it sounds like a big change the substantive code changes aren't that huge.


theduckylittle:
Just to get the cross tagging, this implements #48

klassenjs:
My thought is internal layers should be somehow identified (either by class or by flags of what actions are appropriate) so various UI components can decide how to deal with them (ignore, don't allow to turn off, don't allow zindex change)

klassenjs:
I might be missing something, but I don't see the setting the buffer doing anything on Select Features.  Win/Chrome/index.html

theduckylittle:
Like it's because you switched branches and didn't delete your hash

klassenjs:
Can you change the buffer after you draw or only before?

With a point and a 9mi buffer, I got it into a state where I couldn't draw anything else and the buffer wouldn't clear the screen when exiting the tool.

klassenjs:
There are a lot of gremlins here.

Buffer 100 ft, select a polygon from Drawing Layer, change buffer to 2000ft, go to measure line - measure buffer as closer to 150ft, go back to select features, buffer is set at 0, draw line, line is buffered (at about 100-150ft).  (Note buffer and selected shape also stayed while switching tools).

klassenjs:
(cont) go back to measure tool... all the measurements are buffered on the map.

elil:
In regard to above point, 5. Multi-point buffer, it appears to only buffer the most recent (i.e. last, not first) point.

elil:
This does not preserve (and display) the original input geometry but rather transforms the input geometry to the buffered geometry.  Showing both might be better.

elil:
@klassenjs pertaining to your above comment (not sure how to reply to specific comment), did you enroll in the "How to be an 'average' user" class by me?  That is some crazy testing but good to put it through its paces.

klassenjs:
@elil I haven't, but I was exploring possible failure modes based on the code and past problems.

Responding to the bug reports in #107 since this has been merged.

Issue #81, opened 2017-05-26
Activate (or reactivate) Super Tab when returning to a tool


elil:
When using a tool, let's say Select Features, if you select another tab without hitting "Close" or "Go" (or a different tool), hitting the tool again doesn't bring the tab up.  

Click "Select Features", click on the catalog tab, turn a background layer on or off, click "Select Features", continue staring at the catalog tab and notice that the Super Tab is not activated.  Either click a different tool (Identify for example), or the Super Tab (and optionally "Close" or "Go") to continue.

elil:
Tested as working.  (Assuming reporter closes)

PR #91, opened 2017-05-27
Always hint to open the super tab. refs: #81

theduckylittle:
Added a new UI hint that indicates when a service is "started."

This brings the super tab "to the top" whenever a user clicks to
start a service.


theduckylittle:
refs: #81

PR #93, opened 2017-05-28
Fix for non-es6 ol inclusion.

theduckylittle:
Coordinate display wasn't fully upgraded to ol4/es6 classes during the merge.  This fixes that.


theduckylittle:
@CaitW - I missed this in the PR because I had already ported stuff to the es6-style of openlayers in master.  Just a heads up, future code should use OL4-style ES6 classes.

Issue #55, opened 2017-05-16
Remove feature tool


theduckylittle:
There's no "remove feature" tool for vector layers.

Acceptance criteria:
1. The catalog gets a new tool that indicates a layer can have features removed from it.
2. Clicking on that tool starts a selector tool that lets the user select a feature on the map.
3. After clicking a feature the user is presented with a warning dialog.  Ok removes the features, Cancel closes the dialog.

theduckylittle:
This done got writ. We done gone close this ticket.

Issue #47, opened 2017-05-16
Feature Editing


theduckylittle:
Implement WFS-T layers.

- Geometry editing
- Attribute editing

We'll use TinyOWS as the example WFS-T server but GeoServer should be tested as well.

brentfraser:
Where are we on this?

brentfraser:
One thing I liked about the original vector editing was the performance when dealing with lots of vectors.  The layer with all the features was rendered with WMS, the user identified/selected a feature (or multiple features?) for editing, the geometry of only that feature was retrieved and presented for editing.  Good stuff.

tchaddad:
Is TinyOWS non-functional? If it still works, maybe incorporating it as a placeholder until the new standard is operational is the best path...

theduckylittle:
My worry is that TinyOWS is not-maintained. The current OGC spec only specifies the way to read features and not the way to write or update them. 

I think tagging this to 4.x is likely the best way to go. I don't want this to fall off the list but it's definitely a feature that could use some sponsorship.

jmckenna:
For the record, some much-needed recent activity for TinyOWS: https://github.com/MapServer/tinyows/pull/94

Also, please keep in mind HTTPS support, in your GeoMoose-WFS-T testing (https://github.com/MapServer/tinyows/issues/93).

brentfraser:
So the current TinyOWS (including PR 94) will not work with a https request?

jmckenna:
I think it needs re-testing, and then update the ticket with test results at MapServer/tinyows#93

brentfraser:
see PR #620

PR #101, opened 2017-05-30
Add modify and remove vector tools

theduckylittle:
1. Modify now works on selection shapes and other vector layers. 
2. Remove is now available for  vector layers.

A modify geometry action was added in order to ensure the store was properly updated and not just the openlayer's layer after a feature was finished.


klassenjs:
I don't know what we have for options or how others view it, but the rubber eraser doesn't convey 'Remove All' to me.  If anything, the trash can conveys more 'Remove All' (throw out the whole sheet) while the rubber eraser conveys more rub-out an item.  @theduckylittle, @elil, @CaitW, others, thoughts?

It looks good to me otherwise.

theduckylittle:
I was flailing a bit there.  I guess for me I have the verbs "remove" and "clear."  So "remove" is delete and I think of the trash can.  Clear has less meaning (and maybe limited usefulness?) so "Clearing" means "erasing" and the eraser makes sense.  This is just how my mind was working when I put the CSS down but there's a lot of room for opinion in there.  I even recognize the fallacy of my "clear" association.

Here's a picture of what @klassenjs  is talking about:
image

theduckylittle:
References tickets #55 and #47

elil:
This is tricky.  I don't really like what is proposed but am slightly hesitant to criticize since I haven't thought of anything better.  I'm reviewing some past options to see if it jogs anything.

remove_and_clear_singular_and_plural

remove_feature

remove_all_features

To me the verb is the same (delete, remove, clear, erase, trash, etc).  What is different is the direct object or what is being acted upon (one of all).  

That is what is clear to me in the previous examples, the difference in singular and plural and the difference in ALL or (ONE).  Changing verbs when the action is the same is confusing.   The verb changing made the previous versions worse.  It is the direct object (or the target of the verb) that is changing.    
 
The way I see this is that we can, 
"[verb] (delete, remove, clear, erase, trash, etc) **ONE** Feature/Drawing" 
or we can (exact same verb), 
"[verb] (delete, remove, clear, erase, trash, etc) **ALL**  . Feature**S**/Drawing**S**"

Unless the verb indicates a difference in direct object number I don't see the verb as the way to communicate this.  (destroy, wipe out, abolish, scorched earth vs decimate, reduce, decrement, etc)

So this is an idea, that we are looking at differentiating the direct object and keeping the verb the same.  I still have no ideas on how to communicate that in icons.  I feel okay about the word communication above.  Maybe a red x over a 1 and a red x over a M (for many - yeah that is stretching it), or maybe a red x for all and a red x over a 1 for one.  Or a subtract icon.

theduckylittle:
Subtract (1) vs Trash Can (all) is something I could +1.

klassenjs:
Random thoughts:

Font Awesome has `minus-circle` which would generally match GM2.x.

While, the use of subtract made sense in GM2 since the draw tools had an (almost too small to see) '+' on them, it always seemed a bit out of place to me as a drawing tool.  I personally like the eraser better.

QGIS uses Trash Can to delete selected features (and has only one tool rather than splitting the one and all cases).  Of course, we don't really implement the concept of selected features.

GIMP uses an Eraser to remove what is clicked on.

Most other software seems to not have a GUI representation for delete object and relies on a menu item or keypress after the object is selected.  The closest is the scissors typically used for cut, but cut implies the object can later be pasted which is not the case for us.

elil:
Are you thinking, "-(1)" and "[trash can] (all)"?

What about keeping the verb the same?

"-(1)" and "-(all)" or "[trash can] (1)" and "[trash can] (all)"?

minus

elil:
or eraser (1) and eraser (all)?

klassenjs:
Semantically I like that.  I am not sure how well it would fit and/or translate.  Would an icon for the verb and a 1 or * subscript/superscript superimposed on it work?  If so do we know how to make it?

My personal preference is eraser for the icon but I wouldn't -1 the others.

elil:
I like 1 and all optionally in parenthesis, (1) and (all).  Do we need super/sub-script?  I guess it can be hard to tell where one icon stops and the next begins without some super/sub-script.  

I'm not sure that the * wildcard is going to be readily recognized as all.  

I like the eraser but am not opposed to the other icons.

theduckylittle:
I really don't like having the additional text descriptors and/or super/subscripts for a few reasons:
1. It makes the CSS patten we've got established a lot trickier.
2. It makes the layout of a single icon much trickier in the catalog.  Particularly once more flexible layouts and smaller screen sizes start getting throw in, we'll have issues relating the text and the icon (overflow spacing is real).

@klassenjs put together another ticket but I'm going to make the following changes:
1. Match gm2's "-" circle and trash can to distinguish the two tools.
2. Add confirmation modals to prevent accidentally doing damage.

klassenjs:
npm install failed in a massive way on travis.  All tests passed locally.

elil:
I was thinking simple not fancy.  Put the text in the icon/image [image from above.](https://cloud.githubusercontent.com/assets/5263258/26608005/1d314e3c-454e-11e7-85b6-53ee485a73b3.png)

I do find these tool tips quite descriptive and helpful.

klassenjs:
That's a bit difficult because the icons aren't images/sprites anymore.  They are a single character from Font Awesome rendered by a CSS ::before class over a <span>.

PR #103, opened 2017-05-31
Tweaks to improve the build speed of the JS.

theduckylittle:
Trying to make fire.

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,6 @@
     "test": "tests"
   },
   "devDependencies": {
-    "babel": "~6.5.2",
     "babel-cli": "~6.14.0",
     "babel-core": "~6.14.0",
     "babel-eslint": "^6.1.2",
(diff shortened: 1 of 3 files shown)


klassenjs:
`grunt build` hangs on Win7 in webpack:build-dev at 61% 390/456 build modules.

0% CPU usage, no disk, usage, plenty of free memory.

theduckylittle:
As we're basically ignoring the source maps for "index" right now, maybe we just drop building that.  It saves quite a bit of time on the build-deploy.  That alone might make the process less onerous.

klassenjs:
I am curious how much time and RAM that would save.

It would be cool if the source maps worked in production for debugging, but I have never figured them out (maybe a Firefox thing).  So what I am trying to say is right now I wouldn't notice their absence.

klassenjs:
Note: It looks like UglifyJS uses the source map in deploy to give line numbers in errors/warnings.

theduckylittle:
closing for further research elsewhere.

PR #104, opened 2017-05-31
Add support for AGS/FeatureService layers

theduckylittle:
This implements the ability to do things like "Select" against an ArcGIS REST FeatureService.

If this code "looks good" and tests well then great! But I'd like the opportunity to remove the HEI-Geo server from the example.  This has been setup graciously for our testing but not for wide distribution in the demo.

Some quirks I've noticed:
1. Buffered polygons seem to choke AGS.  Not sure what the issue is but the requests seem fine.
2. I think there may be some quirks with flipping back and through the select and buffer boxes.  I suspect that's more state management quirks that should have some follow up -- please note what you can find as consistently as you can and I'll work on recreating and fixing the issues there.
3. map.jsx is getting horrendously long.  The query functions are in there because they need to know a lot about the map-component and some of it's private members.  I think there may need to be a dedicated clean up project here once we know the rest of the query code is solid.


theduckylittle:
Proof it "works for me": 

image

Select with a point from AGS...
image

Reflexive select:
image

klassenjs:
It appears to work as advertized.  Note: Identify and the Select Features grid are not configured in the mapbook.  Not merging because we need to find a more permanent service for the demo (or not demo it).

klassenjs:
I added some Dakota County layers to help with testing.

This exposed some bugs:
 1) It appears only the first layer defined is used, the layer name is ignored.
 2) The zoom-to seems broken (maybe becuase data is returned in 3857)?

klassenjs:
3) Paging is not enabled, so resulting in only the first 1000 features are ever requested.
4) Pan/zoom does not re-request data for the new extent (although the first view does filter by extent).

klassenjs:
4) Update: It appears to make a new request on zoom out/pan, but not on zoom in.  My guess is it doesn't re-request an region it already thinks it has.  This might be correct behavior, but it looks broken because of 3.

theduckylittle:
1. This is a discussion point.  GeoMoose's data model has a bit of an issue handling more than one "real" vector data layer in the application.  This is due to the way that both E**I, OGC, and "Flat" vector files tend to work.  One URL == One Data Source.  In fact, in GM2 we didn't bother with a "layer" definition.  In GM3 vector "layers" are really just a trick to allow us to do more complicated vector styling.  Only the first "name=" is used as the data source and the rest are "virtual" names for the layers and not alternative sources. If you look at the architecture of OpenLayers, it mirrors this type of pattern.  I'm open to discussion on how to improve the clarity here are my thoughts on moving forward:
  a. Leave it as is and document the hell out of it.  This is probably a lame answer.
  b. I'm apprehensive to setup multiple OL Sources to handle the vector layers.  In otherwords, I don't want to have one vector map-source definition spawn multiple instances of ol.Source.  It really breaks a lot of the assumpting in map in a way that I think makes the code less clear and harder to maintain/test.
  c. Maybe we force the full layer URL to be defined in the vector URL.  I've been avoiding that as the WFS URLs can get to be onerous and right now we've got some nice utilities for glossing over that.

2. Should we force the source to return 4326? We can do that. We have the technology.  But it would probably be easier to change "zoomTo... 4326 ..." to "zoomTo ... 3857 ..." in your templates.

3. Doh! Paging! Again, open minded as to how we handle this.  We could add a <param> that allows for a maximum number of features and also set the "default" to something like 100,000.  See more notes on  point 4 below.

4. This is the OpenLayers "bounding box" strategy.  It assumes that the previous requested all the features it needed to fetch.  Another strategy I've seen in their examples is to use a "tile" based strategy that will populate more detail at different zoom levels.  Ergo, if you populated x=3,y=3,z=3 and then went to x=3,y=3,z=4 there would be another fetch.  It may also serve to solve the paging issue.  (Both solutions may be necessary.)

tchaddad:
Eli suggested you might want access to an ArcServer for testing certain scenarios, but it look like maybe Jim beat me to it.

If you need me to post any test layers let me know.

klassenjs:
1) I think that your assertion that multiple layers within a vector endpoint is uncommon may have been true but is becoming increasingly false.  WFS, ArcGIS FeatureServer, and KML all support multiple layers from one endpoint (although WFS and FeatureServer only support requesting one of them at a time so maybe that is just semantics).  As we are sort of stuck with with OL givith, and the further we get from OLs abstractions the more crazy difficult bugs we end up chasing, multiple map-sources per vector source might be the best workable, but verbose solution.  I almost wonder if we should change the mapbook to better reflect the structure for the vector layers types (like move the layer name, as far as the actual service is concerned into a map-source attribute and have the `<layer>` tag just be internal to GeoMoose).  Does GeoMoose use second part of the path for vector layers?

2. That's fine with me so long as it is documented somewhere.  I wasn't clear if GeoMoose was normalizing it internally to 4326 or not.  Note: The geometries displayed correctly so it somehow picked up they didn't need to be reprojected from 4326 to 3857.

3. Upping the max feature count won't work.  Most ArcGIS FeatureServers I have seen enforce a upper limit of 1000.  The good news is they tell you in the response if they gave a truncated return.  I think they let you make subsequent requests with offset set to fetch the rest of the dataset (in WFS mode I know this needs a relatively recent version of the stack, I don't know if FeatureServer is different.)  Also odd is a max of 1000 records per request seems small enough that it would generate lots of requests for most layers.  From a always rendering a usable map standpoint, using maxscale might also come into play here (maybe with a fallback to a wms/raster tiles path above that maxscale(denom)?

4. I also wonder how many features we can keep in memory in the browser.  We may need a way to tell the user the data is incomplete and they need to zoom in, and a way to forget the results from previous requests.

theduckylittle:
1. NO DOUBT vector services are evolving.  We just have quite a few boxes that provide us with some restrictions.  OL being the primary box.  I'll break out a few more thoughts:
 a. WRT the layer name.  GeoMoose *does* use the first layer name to provide a layerName for the Catalog/SRC/MapSource definition. So the layer name is somewhat important in making the parser and path-names consistent.
 b. I really would like users of vector layers to be able to define "virtual" layers which provide them with the ability to do more complex styling.
 c. Personally, I found having a bunch of "custom" attributes for each vector type to be onerous.  Right now, I find the definition in the mapbook to be a lot cleaner.  And the definitions are more normalized and more documentable right now.

2. This could be a bug.  I got the stuff to work and declared victory. 💯 Historically, we've normalized incoming data to 4326.  The "4326 vs Map Coordinates" debate is an interesting one overall but should probably beshelved to another ticket.  Due to some bugs with MapServer + WFS, those requests are being returned as 4326.  Projections are not a "mess" right now but they are not fully addressed either.  There are comments and warnings and variables saying "Plz pull from the store at some point" and I've got "Make non-3857 projections do-able" as a very important todo item on my list.  (Expect to see an issue regarding how we think projections should be delivered to users shortly...)

3. I was really hoping I could force the server to just return more features.  Win some, lose some. E**I strikes again. A lot of the examples I've seen use the XYZ request strategy.  It may be how they're faking the 1,000 feature issue.

4. I suspect the XYZ strategy also help with memory issues.  At least some representation of the features can be tossed when the tiles are no longer valid AND each tile will, ostensibly, have fewer features.  JavaScript applications have grown grotesquely memory hungry over the years.  In no small part because the VMs that run JavaScript have had more memory to use and optimize.  Not sure there will end up being a good answer besides "Don't hurt yourself."

I think the follow up for this is:
A. *Try* the XYZ loading strategy.
B. Make a decision on how we want to handle projections from service (Always store in 4326? Always in Map Coordinates?).

theduckylittle:
Also, I think we'll ticket the "Notify users that more data is available" but I'm not 100% sure we'll get it into 3.0.0.

theduckylittle:
Issue #108  has been opened to discuss feature projection issues that were mentioned here.

tchaddad:
The ArcServer admin can allow the server to respond to > the default number of features, but not the client, but it has to be deliberately set.

theduckylittle:
Pushed updates:
a. XYZ strategy is now enabled.
b. Fixed the bounding-box issues for dakota streets, three-fold improvement: fixed the path reference bug, fixed the bounding boxes to ensure they are in the right projection, ensured AGS layers get a "boundedBy" property.

klassenjs:
XYZ strategy is better but doesn't solve the 1000 feature limit 100%.  It doesn't appear to re-request on zoom in.

screenshot from 2017-06-06 07-53-26


After reload page (hashtrack saves state)

screenshot from 2017-06-06 07-55-15

klassenjs:
Select and zoom-to on streets appears to work.

theduckylittle:
I've flushed this out according to this documentation:
 - http://resources.arcgis.com/en/help/rest/apiref/fsquery.html

Essentially there is now a minimum of two requests per "tile" possibly more:
* For each XYZ combination get the list of object IDs.
    * For each batch of 500 IDs make a query by ID.
      * Add each 500 feature batch to the source.

I picked 500 as it is "half" the default maximum as a bit of a precaution against a GIS admin who might think "OMG, setting this to half".

theduckylittle:
Research from @klassenjs shows that E**I was throwing a fit for QUERY_STRINGs > 2048 characters.  New batching code accounts for this and works around it.

klassenjs:
Damn it's slow for the full data set, but it appears to work.

I think the last thing left is deciding what we want in the demo.  All three?, just Dakota Co?, leave them in but commented out? ...

klassenjs:
Found one more issue (and it may be bigger than this PR): identify doesn't seem to work against line features (point never intersects line, unless it is exact within round-off error.)

klassenjs:
The above commit though should get the mapbook and app.js into a state good enough for the demo.  I kept the two Dakota County layers, but commented out Streets (it is large, can slow the browser, and probably isn't that easy on their server, but a good example and test case for more complex ags-vector layers).  And I moved them into Overviews/County Layers in the catalog.

PR #105, opened 2017-05-31
Attempt to fix travis-ci builds

klassenjs:
It looks like travis-ci updated their default node from v7.10.0 to v8.0.0.  This means the iltorb module is not available as a binary and needs to be built from source.  This in turn requires a C++11 compiler which in turn requires update the distro from Ubuntu Precise to Trusty.


klassenjs:
... but that still has build errors so try forcing node 7?

Issue #107, opened 2017-05-31
Client Side Buffer Oddities


klassenjs:
Continuing the discussion here from PR #88 since it has been merged:

@elil  Using 8487451 on https://demo.geomoose.org/3.0/ I did select features, draw polygon, change buffer to 500 ft, go and it appears to show the original polygon, the buffer, and the selected parcels.  Is this working as you think it should?

![screenshot from 2017-05-31 10 30 52](https://cloud.githubusercontent.com/assets/165622/26640109/5faa3b2c-45ec-11e7-961d-5fc9fbfe79ae.png)

I see multi-point is only using the last point.  Select Features, Draw Multi-Point, set buffer 500ft, go:
![screenshot from 2017-05-31 10 35 52](https://cloud.githubusercontent.com/assets/165622/26640292/f1c86664-45ec-11e7-9e93-ddcfddce2f37.png)

I also see that the buffer setting stays set to what it was when re-entering the Select Features tool, but the form elements are set to 0ft.  Changing the buffer distance form element updates the buffer to the value displayed.
![screenshot from 2017-05-31 10 39 42](https://cloud.githubusercontent.com/assets/165622/26640495/7b121802-45ed-11e7-990c-5dc69c179919.png)

elil:
Yes, showing original input geometry is working as I think it should.  

I note the same Draw Multi-Point issue as you with only bufferng the last point.  Also, Draw Multi-Point does not remove the previous drawing.  

![multipoint_doesnt_clear_previous_drawing](https://cloud.githubusercontent.com/assets/5263258/26642742/bb61360e-45e3-11e7-91cb-8b0ed6de2d77.png)

We also need to decide when to clear the previous selection see below:
![when_to_clear_previous_selection](https://cloud.githubusercontent.com/assets/5263258/26642820/1037b8ba-45e4-11e7-8970-90a67312f8a0.png)

klassenjs:
#110 Fixes what I saw so I am going to merge that (under the strategy that it is at least better than it was).
@elil, can you verify if this fixes what you saw or if another PR is needed?

theduckylittle:
I would <3 a summary list of what quirks are still happening.

elil:
Summary:
- [X] show original geometry
- [X] Multi Point uses all points
- [X] Multi Point removes previous geometry
- [X] Re-entering the Select Features tool, form elements and buffer match
[...]

PR #110, opened 2017-06-05
Fixing Buffering Quirks

theduckylittle:
1. Fixed quirks with the buffer input.  
  a. The buffer "live update" now works more closely with the state.
  b. That distance updates properly.
2. Fixed issue with multi-point/multi-geometry selections.  They are now buffered then unioned.


theduckylittle:
Refs: #107

Issue #94, opened 2017-05-29
Activate focus of search box when search icon is clicked


elil:
If you click the Search Parcels icon and blindly start typing the text does not go in the search box.  It would be better if that text went into the box.  If there were more than one box, it could perhaps be configurable which box was the default (although ideally search would be effective enough that only one box is needed).  

Related, #61 since a common use would be to click Search Parcels, blindly type your search without clicking in any boxes, hit enter, see results.

klassenjs:
@elil Please Review.

elil:
Yes, this works and combined with #61 now eliminates lots of intermediate clicking.  

(I think that there is a lot more clicking to get rid of but that is a separate matter.)

Issue #61, opened 2017-05-16
Enter key does not make services "Go"


elil:
Right now the enter key does not appear to do anything.  

When someone does a search or other interaction, the enter key should trigger the "Go" button.

klassenjs:
@elil Please review.

elil:
Enter key now makes it go.  Escape is a nice addition as well.

PR #119, opened 2017-06-07
Added keyboard short-cuts.

theduckylittle:
1. ENTER in an input is the same as clicking "Go".
2. ESCAPE in an input is the same as clicking "Close".
3. The first input element will be selected after render.

Refs: #61 , #94

### DIFF
--- a/src/gm3/components/serviceManager.jsx
+++ b/src/gm3/components/serviceManager.jsx
@@ -418,6 +418,37 @@ class ServiceManager extends Component {
         }
     }
 
+    /** Function to handle bashing "Enter" and causing
+     *  the service form to submit.
+     *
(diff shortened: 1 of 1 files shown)


klassenjs:
There are some state bugs here (possibly unrelated).
* Escape closes Select, but doesn't end Drawing or clear what was drawn,
* Return does go, but going back to select, need to switch drawing tools to be able to draw again.

theduckylittle:
@klassenjs -- those are separate bugs.  The first one is a part of the discussion @elil brought up regarding some indeterminate states between selection/results.  Your second point is the same behaviour regardless of ENTER vs Go.  We can file it as a bug if we want. Or the user can click Draw.

PR #121, opened 2017-06-07
Fixed the failing template issues.

theduckylittle:
1. Removed the missing template from the mapbook
2. Reworked the getTemplate promise to ALWAYS resolve, it now resolves to a blank template on failure to load.


klassenjs:
If I use the old mapbook with the new code, identify returns nothing (with City and County + Parcels on).

theduckylittle:
Found the issue, without handling the failure condition expressly, the promise would reject.

Issue #84, opened 2017-05-27
Select Features using "Select from Drawing and Markup" allows for selectiong mutiple drawings but only uses the last one


elil:
Make various drawings and markup.

Select Features... "Select from Drawing and Markup," press and hold shift key, click and select multiple drawings, hit go.  The selection returned in based only on the last selected drawing.  Selections can also be of multiple geometries types (may be fine but may need to take that into account).  

Selection should either be for all selected drawings or selection should not allow multiple selections or there should be an error message to only use one selection.  

Measure and other aspects are probably a separate issue, or maybe this issue here should be a sub-issue of the general issue of handling multiple drawing selections.  In measure, a user can select two drawings and it does not sum their area (also, the user is not prevented from selecting different geometry types).

klassenjs:
Cool, I had no idea shift key would allow multiple selections.

klassenjs:
@elil Please Review.

As I said in the PR, I could see multi-select being a future enhancement, but since this isn't a regression from 2.x and in the short term I'd rather have a stable 3.0.0 release rather than waiting to solve the hard problem of how to make the tools multi-geometry aware.

elil:
Is #122 supposed to be in the demo now?  I can still select multiple geometries.

klassenjs:
The demo works for me. I see it at cfcb073 which includes this fix.

Try clearing browser cache.  I have found sometimes it doesn't pickup on the new geomoose.min.js.

PR #122, opened 2017-06-07
Turn off multi-select.

theduckylittle:
This was causing a litany of bugs.

refs: #84


klassenjs:
I could see potentially adding multi-select functionality back as an enhancement in the future and making the necessary things mulch-geometry aware, but I think this is a good solution for now in terms of reaching a stable 3.0.0 release in a finite amount of time.

PR #123, opened 2017-06-07
Fix for tab overflow scrolling bug.

theduckylittle:
Scrolling tabs was causing the tab headers to scroll off.


klassenjs:
Before:
screenshot from 2017-06-07 10-35-04

After:
screenshot from 2017-06-07 10-35-21

Issue #127, opened 2017-06-07
Metadata Layer Tool


klassenjs:
Used to have one, probably should still have one.

PR #131, opened 2017-06-09
Return of the metadata tool!

theduckylittle:
Refs: #127

### DIFF
--- a/examples/desktop/mapbook.xml
+++ b/examples/desktop/mapbook.xml
@@ -485,7 +485,9 @@
                draw-modify="true" draw-remove="true" />
 
 		<layer title="Minnesota Places"  activate="true" src="minnesota_places/gm:minnesota_places" fade="false" unfade="false" metadata="false" legend="false" draw-point="false" draw-line="false" draw-polygon="true" edit-attributes="true" edit-shape="true" remove-feature="true" remove-all-features="true" buffer="true"></layer>
-		<layer title="Vector Parcels"  activate="true" src="vector-parcels/ms:parcels" fade="false" unfade="false" metadata="false" legend="false" draw-point="false" draw-line="false" draw-polygon="true" edit-attributes="true" edit-shape="true" remove-feature="true" remove-all-features="true" buffer="true"></layer>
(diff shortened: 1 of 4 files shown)


examples/desktop/mapbook.xml
```diff
@@ -485,7 +485,9 @@
                draw-modify="true" draw-remove="true" />
 
 		<layer title="Minnesota Places"  activate="true" src="minnesota_places/gm:minnesota_places" fade="false" unfade="false" metadata="false" legend="false" draw-point="false" draw-line="false" draw-polygon="true" edit-attributes="true" edit-shape="true" remove-feature="true" remove-all-features="true" buffer="true"></layer>
-		<layer title="Vector Parcels"  activate="true" src="vector-parcels/ms:parcels" fade="false" unfade="false" metadata="false" legend="false" draw-point="false" draw-line="false" draw-polygon="true" edit-attributes="true" edit-shape="true" remove-feature="true" remove-all-features="true" buffer="true"></layer>
+		<layer title="Vector Parcels" src="vector-parcels/ms:parcels">
+            <metadata>http://google.com</metadata>
```
elil: A perhaps more logical link would be the help page for metadata.

theduckylittle:
Any and all suggestions are welcome @elil

elil:
This is of course crosses the 2.9/3.0 divide but this link has the spirit of metadata documentation, https://docs.geomoose.org/2.9/docs/mapbook.html?highlight=metadata#catalog.  Or we could link to the actual metadata/license page for the data, https://github.com/geomoose/geomoose/blob/master/maps/demo/parcels/LICENSE.  We could also make that into html and host it on the geomoose.org domain.  

All this is operating under the "metadata" concept.  If it were a "link" concept, than that is entirely different and google might be a good link example.  

Do we want to think "metadata" (a link that is only about the specific data) or "link" (any valid url)?

klassenjs:
The intent is a link to a valid url that represents metadata for the specified layer.  The implementation is rather agnostic to the contents of the link.

Issue #89, opened 2017-05-27
HashTrack and "Built-in" Layers


theduckylittle:
Two different types of "smart" layers have been added to the application:
1. "results/*" there are two different results layers.  One for display the results of a query (and honouring the client side filters for it) and one for hovering over the results in order to see a 'selected' geometry on the map.
2. "selection/selection" While writing the buffer tool it was important to preview what the user was going to get when they selected applied a buffer. This required having two "seleciton" layers. The first selection layer is built-in to the OpenLayers map and is a consistent target for the OpenLayer's controls to use. the second selection layer is the map-source "selection" with the layer "selection" which is managed by the map to preview the actual selection features as the user gets them.  This is added with PR #88 .

However, when a user decides to remove "results/" or "selection/" layers from the hash then when the services render they can be missing those layers and not see the highlighted results or the selection preview.  i have a couple of thoughts on this:

1. The layers could be forced on anytime they get new data or something is expecting them to be visible.
   This seems like a reasonable "hack" but I do feel like these layers can ALWAYS be on and it will not effect the user experience with how OpenLayers now handles vector layers (that is OL3+ vectorizes them).
2. Add a new "hide from tracking" or "alwaysOn" status for the layers.  
   I actually feel this is a reasonable thing to do.  Some layers are internal, they should always be on and they're lynch pins for good functionality.  
3. Put these layers BACK into the Mapbook and let users add them to the catalog.
   This certainly is an approach and offers a lot of flexibility.  There could even be some CONFIG variables used to set these with reasonable global states and everything.   *BUT* I honestly believe there were some times where we really put effort into maximum flexibility on features and lost consistency from a user's perspective.  Yes, a savvy administrator had more tricks available to them but we had a harder time managing state and in the new version I think we'd have a more difficult time doing testing.  The corner cases start growing out of control.

theduckylittle:
I'd like to implement option 2.

klassenjs:
+1 for Option 2.  I like the flags approach, maybe with the option for more fine grained flags in the future.

elil:
I agree with above, point 3. flexibility and consistency.  Flexibility is very powerful and something that makes GeoMoose fairly unique.  It is also something that complicates our testing and expands our corner cases and can lead to whack a mole.  I'm not sure of the balance to strike.  One complex way would be a mapbook validation tool that identifies conflicting mapbook configurations. But then we would need to write rules for valid mapbook configuration.  

Also, #95, about hovering might more properly belong here.

[...]

PR #132, opened 2017-06-09
Add support for `always-on` layer.

theduckylittle:
This implementaiton uses a "param" of "always-on",
this will let administrators specify their own layers
as never being able to be turned off.

Refs: #89

### DIFF
--- a/src/gm3/application.jsx
+++ b/src/gm3/application.jsx
@@ -131,9 +131,10 @@ class Application {
             queryable: false,
             refresh: null,
             layers: [],
-            params: {
-                'always-on': true,
+            options: {
+                'always-on' : true,
             },
+            params: {},
             // stupid high z-index to ensure results are
             //  on top of everything else.
             zIndex: 200001,
@@ -185,9 +186,10 @@ class Application {
             queryable: false,
             refresh: null,
             layers: [],
-            params: {
+            options: {
                 'always-on': true,
             },
+            params: {},
             // stupid high z-index to ensure results are
             //  on top of everything else.
             zIndex: 200002,
--- a/src/gm3/application.jsx
+++ b/src/gm3/application.jsx
@@ -132,7 +132,7 @@ class Application {
             refresh: null,
             layers: [],
             options: {
-                'always-on' : true,
+                'always-on': true,
             },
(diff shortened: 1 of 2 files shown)


klassenjs:
Is always-on accessible from the mapbook?

theduckylittle:
Yes. But the built-in layers have it set by default.

Mapbook syntax:
<Param name="always-on">true</param>

klassenjs:
I can't get the mapbook part to work.  I'm trying to set `always-on` on the Drawing and Markup layer as a test.
I have tried map-source->param, and map-source->layer->param, (and earlier map-source['always-on'], and  map-source->layer['always-on]).

klassenjs:
For the record: I see it working properly with the internal layers.

theduckylittle:
The catalog doesn't honour it as we were trying to fix the trackers.  Adding a "lock" to the catalog could get ... interesting particularly if you had mixed always-on and non always-on layers.

klassenjs:
My test was if it removed from tracking, I didn't think to test the catalog.

theduckylittle:
I have updated comments in the ticket.  But the syntax I suggest was wrong it's `<param name="always-on" value="true"/>` and there is an update to also check the value with `parseBoolean` if a straight boolean test does not pass.

klassenjs:
Does using param this way confuse what gets sent to configure the layer in OL/WMS?  I know it was more or less a pass through in the old days, don't know about in GM3.

theduckylittle:
Yeah, it still is basically a pass through for the WMS-style layers.  We could add another section like "option" or "conf" if we really need to. The scope of the ticket was to fix this for internal layers, so we could um... not document this as recommended for the mapbook.

klassenjs:
Could move it to yet another attribute too.  But the confusion of <param> thing bothers me.

klassenjs:
No mapbook support and fixing the docs would be valid for now too. But I don't like overloading `param`.

theduckylittle:
I've moved the "always-on" from "params" to "options".  If at some point we decide to move that to the mapbook we can.  Either by implementing a "<option>" parser or by mapping attributes to the options object.

Issue #128, opened 2017-06-07
Drawers in Toolbar


klassenjs:

theduckylittle:
@klassenjs  -- this more work now?

klassenjs:
Yep.  Tested on FF, Chrome W10 desktop and tablet (touch) modes and Chrome Android.

PR #136, opened 2017-06-12
We're bringing the drawers back...

theduckylittle:
And all I've got is timberlake lyrics going on in the back of my head...

This adds drawers back to the Toolbar.


theduckylittle:
refs: #128

elil:
How do we decide between putting options in drawers or in the tab?  Search now has drawers.  Measure gives tab options for point, line, polygon, select.  Those could be drawers.

klassenjs:
The things in the toolbar are separate tools.

The different drawing types (point/line/polygon/select from layer) are inputs to the same tool.

Issue #137, opened 2017-06-14
IE is DOA in Master


theduckylittle:
This is just a nice friendly place to remind all the devs that IE is real, we should work on IE, and that we can't just hide from it.

Cheers,
- The real world

theduckylittle:
Yay! Works now!

PR #138, opened 2017-06-15
Fix for IE iterator bug.

theduckylittle:
IE treats the returns form getElementsByTagName as a "collection" instead of as a basic array and therefore support for ES6 "of" operator doesn't transpile correctly.  This fixes the transgression.


theduckylittle:
refs: #137

klassenjs:
I haven't fully tested IE11 or Edge but both generally appear to work with this patch.

PR #141, opened 2017-06-16
More coverage

theduckylittle:
Major push to get more tests integrated into the UI.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
   "homepage": "http://www.geomoose.org",
   "bugs": "https://github.com/geomoose/gm3/issues",
   "scripts": {
-    "test": "node node_modules/jest/bin/jest.js --coverage tests",
+    "test": "node node_modules/jest/bin/jest.js --coverage tests/",
     "prepublish": "grunt build",
     "start": "grunt serve",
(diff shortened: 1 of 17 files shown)


theduckylittle:
@klassenjs This is the promised testing update I threatened to write.

klassenjs:
Travis says toolbar is broken.

theduckylittle:
I think that's because of a mid run recommit. If you look up in the list, Travis finished fine and the change wouldn't have effected test output.

klassenjs:
It did test ok locally.

I'm not sure about the explanation.  I haven't known travis to update the files mid test.  The log on travis shows it did complete, but with one a failed test.  It isn't like it was interrupted.

I wonder if it has to do with node versions.

theduckylittle:
Yes but look at the difference between commits. The first one passed and seconds later I pushed another that corrected a file *that does not affect the tests.* So there's some weirdness going on.

klassenjs:
Sorry, it has been a busy day.  Ok, I see that.... If it is a legit problem it will show up on the next one.

klassenjs:
Tests passed in master.

PR #144, opened 2017-06-20
Cleanup of mapbook pre-release.

theduckylittle:
1. fixed the scale bar layer names.
2. removed a pile of cruft.
3. The metadata links for parcels now points to the
   LICENSE on github for parcels.

### DIFF
--- a/examples/desktop/mapbook.xml
+++ b/examples/desktop/mapbook.xml
@@ -23,97 +23,6 @@
             to the map.
             ]]></legend>
         </layer>
-
-		<attribute name="title" type="user" default-value="" label="Feature Label:"/>
-		<attribute name="line_color" type="color" default-value="#ff0000" label="Stroke Color:"/>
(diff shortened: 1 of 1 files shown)


klassenjs:
How are we interpreting <mapbook version="...">?  Is anything looking at it at all or is it effectively just documentation?  Ideally, we'd use it to identify versions of the mapbook not compatible with the application (and/pr interpret old mapbooks in new app versions).  Mostly, it seems odd to me to be 3.0.0 when symver implies all 3.0.anything should be compatible and any 3.x should be compatible with any app 3.y where y >= x.  Depending on how we actually interpret this though it might not matter.

(Yes this probably should've been an email, but email is hard right now and github just works.)

theduckylittle:
Uh... I'm not entirely certain what you're asking...

klassenjs:
I'm +1 to merge either way.

PR #158, opened 2017-06-22
Major update to vector layer handling.

theduckylittle:
1. typename is no longer inferred for WFS layers,
   it's a required parameter.
2. ags-vector layers must now specify the *full* path
   instead of inferring it from the layer name.
3. `layer`s in vector layers are now basically for styling.
   They work with the list of features that have been returned
   from the vector source.  The mapbook hsa been updated to show
   "expensive" parcels as an example.
4. Properties are now transformed when injested from vector sources
   as well as from queries.  This allows the user to force the datatype
   of a feature property.
5. The mapbook and app.js has also been updated in the desktop example
   to use "common names" instead of direct layer names to prevent
   confusion and conflation.

### DIFF
--- a/examples/desktop/app.js
+++ b/examples/desktop/app.js
@@ -66,9 +66,9 @@ app.loadMapbook({url: 'mapbook.xml'}).then(function() {
     app.registerService('search', SearchService);
     app.registerService('select', SelectService, {
         queryLayers: [
-            {value: 'vector-parcels/ms:parcels', label: 'Parcels'},
-            //{value: 'ags-vector-dc20/20', label: 'Dakota County Streets'},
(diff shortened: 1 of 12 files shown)


theduckylittle:
@klassenjs I've updated this from WIP to ready.

theduckylittle:
Changes made in the second phase:
1. Moved the storage of "GeoJSON"/pure-vector layer features
   from the first layer to the map-source itself.  This makes it
   consistent with other vector layers.
2. Updated all of the related tools for working with vector layers.
3. Updated all of the related actions.
4. Wrote a new set of tests which ensures that all of the reducer
   changes, in fact, work as predicted.

Issue #160, opened 2017-06-23
Not all backgrounds display


elil:
On "OpenStreetMap - Black and White" nothing displays and I get this error in firebug: Loading mixed (insecure) display content "http://a.tiles.wmflabs.org/bw-mapnik/12/985/1479.png" on a secure page

It looks like this might be available in [https](https://a.tiles.wmflabs.org/bw-mapnik/12/985/1479.png) as well but the certificate is not entirely valid. 

On "Reprojected Regional Aerial" nothing displays and I don't see anything in firebug.  Print logs errors in firebug.

USGS DOQs and Topo Quads are annoyingly slow.  They also don't print on the first try.  Returning to them later and trying again succeeds in printing.  Also, sometime when switching between these two, the wrong one displays.  

ArcGIS 9.3 Rest Example does not print.  Firebug logs some errors while printing. 

Bing Roads and Aerials do not display.  Also, does not print.  Print logs errors in firebug.

klassenjs:
I know the B&W layer used to work (albeit with a warning).  Is this a regression caused by #148?

klassenjs:
The print preview dialog is an interesting work around.  For "reasons" (as Dan would say) it is difficult to detect when all the layers have finished rendering.  By waiting for the user to click print, it is a work around to have the user verify the map is right before generating the PDF.  The print preview feature was a pleasent side effect.

klassenjs:
Are all the background layers OK to allow to print (license/terms of use)?  If not we probably need a layer flag for don't allow print and if it is set throw a warning to the user when they start a print.

theduckylittle:
Lemme break this down:

1. Regional aerial: just fixed it.  There was a projections support issue. The MS4W demo data package will need updated and I'm not sure if GeoMoose.org is updating the data repo automatically.
2. Everything else is in PR #162.

theduckylittle:
PR's are ready, review should happen shortly.

elil:
On "OpenStreetMap - Black and White" now displays.  It does not print but I think that is by design.  If that is by design, there should be some message to the effect, "blah layer does not print."

"Reprojected Regional Aerial" now displays and prints.

USGS DOQs and Topo Quads are still a little slow.  When switching between these two, the wrong one still displays sometimes.

ArcGIS 9.3 Rest Example now prints. 

Bing Roads and Aerials have been removed so are no longer relevant.

theduckylittle:
[...]

PR #162, opened 2017-06-24
Major bug fixes for printing

theduckylittle:
1. Added a `printable` attributes to map-sources.
   This defaults to true but is set to false for the WMF Labs tiles
   which do not support cross origin printing. Note: this may work
   with some apache header settings but I haven't tested it yet
   and not wanting to enforce web-server rules I made the decision to
   make this just not-printable.

2. Commented out Bing layers.
   These don't work without a key from Bing. When the tutorials
   are published they will include how to add back in Bing layers.

3. Made specifying the cross-origin parameter more clear.
   OpenLayers will set a CORS header when given a little information,
   the ol.source.OSM class sets this to anonymous by default but I
   didn't want to map any additional layer type when XYZ layers will
   work fine with the additional parameter.  This can be further
   built-out to include layers which require authorization.

4. Similar to the issues with OSM layers, this adds cross-origin
support to AGS raster layers.

5. There were tests failing due to "params" being undefined
for map-source's.  This ensures other tests can rely on
map-source.params existing.  It also sets a couple of other
handy defaults.

refs: #160

### DIFF
--- a/examples/desktop/mapbook.xml
+++ b/examples/desktop/mapbook.xml
@@ -200,6 +200,7 @@
 		<url>https://services.arcgisonline.com/ArcGIS/rest/services/NatGeo_World_Map/MapServer</url>
 		<layer name="NatGeo_World_Map"/>
 		<param name="FORMAT" value="png"/>
+        <param name="cross-origin" value="anonymous"/>
 	</map-source>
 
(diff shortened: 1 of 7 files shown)


theduckylittle:
5. Fixed issues with the black-and-white layers + HTTPS.

PR #167, opened 2017-06-26
Clean up of tools vs icons and icon definitions.

theduckylittle:
We had icon definitions all over the place and they weren't
consistently defined. This cleans up those definitions.  There
are now "icon" classes which can be used anywhere to provide
a clearer set of pictograms for relating to verbs and nouns in the
application.


theduckylittle:
refs: #168

src/gm3/components/layers/wms.js
```diff
@@ -46,7 +46,10 @@ function defineSource(mapSource) {
         url: mapSource.urls[0],
         // This is a carry over from previous generations behaviour.
         ratio: 1.0,
-        params: Object.assign({'LAYERS': layers.join(',')}, mapSource.params),
+        params: Object.assign({
+            'VERSION': '1.1.0',
```
klassenjs: Version 1.1.0 looks odd.  I've seen 1.1.1 and 1.3.0 in the wild.  It looks like OL4 wants to use 1.3.0 by default.

http://openlayers.org/en/latest/apidoc/ol.source.ImageWMS.html

klassenjs: WMS version is really something that should be settable in the mapbook.  A server doesn't have to advertise/accept all the versions.  I don't know if it makes more sense to set this up as a wms vs wms13 layer type or via a parameter.

theduckylittle:
I've fixed the versions. But to address @klassenjs 's concern re: versions, the version specified in `<param>` will override the default version.

PR #169, opened 2017-06-26
Projection fixes + Service Display Enhancements

theduckylittle:
1. Ensures that all query contents are converted to
   map coordinates. Before it was possible to get either
   map coordinates or 4326.
2. Stripped projection specification from examples.
3. Added support for "header" and "footer" in select service.
4. Added new generic header that informs the user the number
   of features found, how many layers were queried, and
   let's them zoom to the extent of the results.
5. Reconfigured select-parcels to just spit out a message
   saying results are in the grid.


theduckylittle:
PR #167 should be pulled in first.  I had to branch this off of it in order to prevent a really ugly merge later.

Issue #140, opened 2017-06-16
Add features from file (to Drawing and Markup)


elil:
Adding features from file to the Drawing and Markup works in the case of three lines totally 12 vertices.  I tried on larger files (the parcels, etc) and it did not appear to work.  It also gave no appearance of error or anything else.  If a file doesn't load, there should be some message, not silence.  

As a side commentary, the description in the Catalog seems a little verbose: "The Drawing and Markup layer can be used to add user defined shapes to the map. "  Maybe that's appropriate in the demo though.

theduckylittle:
Which file did you use Eli? Can you post a URL to an example GeoJson or KML file?

elil:
These came from the parcels (in the 2.# line).  I converted them with QGIS 2.18.3 using the simple, right click shp in catalog, click save as...  So OGR underneath but I didn't bother to do it directly in OGR which sometimes produces better results.  

[test_parcels _kml.zip](https://github.com/geomoose/gm3/files/1081109/test_parcels._kml.zip)
[test_parcels_geojson.zip](https://github.com/geomoose/gm3/files/1081110/test_parcels_geojson.zip)

theduckylittle:
I found out why...  the files you're trying to upload are in EPSG:26915.  The Upload assumes files are in 4326 and then projects that to the map coordinates.  It's noting to do with large files.  I fixed the projection issue and the files loaded up 100% fine.

klassenjs:
So my read on this is it technically "works" but should inform the user when something went wrong rather than giving the appearance of silently failing.

theduckylittle:
This is a specific case of garbage-in-garbage-out.
Dialog improvements need to be done but there's a lot of gotchas that make it a bigger project.  I have some thoughts on it but I'm always willing to examine someone elses PR. LOL

elil:
I see that test_parcels_geojson.zip is EPSG:26915.  

test_parcels _kml.zip is 4326 (values like -93.0461247360058,44.5435302527502) and exhibits this problem. 

I tested with geojson in 4326 [test_parcels_4326.zip](https://github.com/geomoose/gm3/files/1090292/test_parcels_4326.zip) which seems to work.  It may lead to another enhancement issue asking that the outline of polygons not be the same color as the fill.  It makes adjacent polygons very difficult to use.

theduckylittle:
[...]

PR #170, opened 2017-06-26
Fix for the "Eli Bug"

theduckylittle:
Updated the styling to be less ... gross and opaque.

@elil I was able to fix this without going "upstream".  It just requires a little different syntax.


elil:
I think that this is referencing the tangential issues in #140

theduckylittle:
Yup. It is.

PR #171, opened 2017-06-26
Adding more helpful text.

theduckylittle:
1. Added messages to the Visible and Favorites tabs
   explaining what they are.
2. Used the new "info-box" style for the info during the empty
   super tab.
3. Added help text to the print dialog for when there are
   unprintable layers.

### DIFF
--- a/src/gm3/components/favorites.jsx
+++ b/src/gm3/components/favorites.jsx
@@ -61,6 +61,11 @@ export class FavoriteLayers extends Catalog {
 
         return (
             <div className="catalog favorites flat">
+                <div className="info-box">
+                This tab lists all layers marked as "Favorites".  Favorite layers
+                can be selected by clicking the <i className="icon favorite"></i> icon
+                next to the layer name.
+                </div>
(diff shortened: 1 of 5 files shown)


theduckylittle:
refs: #160 -- adds help text for when there are missing layers from the print.

theduckylittle:
Also, I'm aware there is an icon that won't render.  That'll magically pop in when #167 is merged.

klassenjs:
I feel this is useful the first time and maybe as a reminder, but that it takes up a lot of screen real estate after the user is acquainted with the app.

theduckylittle:
Having obvious help, IMO, is better than leaving folks stranded.  If an admin wants to turn off the help text in those two places they can do so in CSS:
```css
.visible-layers .info-box, .favorites .info-box {
 display: none;
}
```

klassenjs:
I wasn't thinking admin turning it off.  I was thinking either the user can hide it.  I suppose if the list is long it will be scrollable anyway so probably not a big deal.

PR #174, opened 2017-06-26
Fix for zombie queries.

theduckylittle:
The "keep only one" setting was not deleting the old
queries in the background.  That was causing the memory
usage to explode over repeated queries.

### DIFF
--- a/src/gm3/reducers/query.js
+++ b/src/gm3/reducers/query.js
@@ -114,10 +114,15 @@ export default function queryReducer(state = default_query, action) {
 
             // when running in single query mode the current query should
             //  be the only one displayed.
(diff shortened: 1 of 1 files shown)


klassenjs:
I can't tell that this fixed anything (I was not seeing the leak), but it doesn't look like it broke anything either.

theduckylittle:
You'd need to check the state tree to verify.  Basically, run a few queries then in the console:
```javascript
app.store.getState().query
```

And you'd see a pile of query results in memory. But because `order` only every showed the most recent, the user could never remove those previous queries from memory.

klassenjs:
Ok, I'll take your word for it.  I was looking at about:memory and top.

PR #181, opened 2017-06-30
How to style the header.

theduckylittle:
Docs abound.


theduckylittle:
Evil self merge to get docs ready.

PR #190, opened 2017-07-20
CSS Tweaks and new LayersList Control

theduckylittle:
1. Added new `layers-list` type select that can be used
   to generate a list of layers based on template filters.
2. Cleaned up and enhanced some of the default desktop app CSS.

### DIFF
--- a/examples/desktop/app.css
+++ b/examples/desktop/app.css
@@ -142,6 +142,13 @@ body {
     }
 }
 
+.feature-class {
+    background-color: #CFE7B3;
+    padding: 2px;
+    font-weight: bold;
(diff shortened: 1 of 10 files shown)


klassenjs:
So, am I getting this right?

Any layer with a select template defined (potentially in the mapbook) is now automatically included in the "Query Layer" drop down, instead of another repetitive section in app.js that was both very unDRY and made it very difficult to use different mapbooks with the same app.js?

theduckylittle:
Yes, @klassenjs , yes. I fixed your biggest objection to the new query service configuration.  It is also possible to override those settings on an application-by-application basis.

PR #196, opened 2017-08-01
Added support for WFS + GeoJSON

theduckylittle:
1. Added the ability to load a map using the GeoJSON return.
2. Updated queries to switch between XML and JSON.

To have WFS queries return GeoJSON set the outputFormat parameter
to 'application/json'.  E.G.:

```xml
<map-source ...>
    <param name="outputFormat" value="application/json" />
</map-source>
```

### DIFF
--- a/src/gm3/components/layers/vector.js
+++ b/src/gm3/components/layers/vector.js
@@ -48,11 +48,12 @@ function defineSource(mapSource) {
         let format = GML2Format;
         let output_format = 'text/xml; subtype=gml/2.1.2';
 
-        if(mapSource.params.outputFormat
-          && mapSource.params.outputFormat.indexOf('json') >= 0) {
+        if(mapSource.params.outputFormat) {
+            output_format = mapSource.params.outputFormat;
+        }
 
+        if(output_format.toLowerCase().indexOf('json') >= 0) {
             format = GeoJSONFormat;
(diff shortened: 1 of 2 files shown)


src/gm3/components/layers/vector.js
```diff
@@ -45,8 +45,18 @@ import getStyleFunction from 'mapbox-to-ol-style';
...
     if(mapSource.type === 'wfs') {
         // add a wfs type source
+        let format = GML2Format;
+        let output_format = 'text/xml; subtype=gml/2.1.2';
+
+        if(mapSource.params.outputFormat
+          && mapSource.params.outputFormat.indexOf('json') >= 0) {
+
+            format = GeoJSONFormat;
+            output_format = 'application/json';
```
klassenjs: I agree that sniffing for 'json' in the outputFormat param is a good enough way to detect to use the GeoJSONFormat over GML, but why not `output_format = mapSource.params.outputFormat;`  Then people can make it work for servers expecting something else.  Technically, you need to give the server back what it offers in GetCapabilities (and as far as I can tell there isn't one single best answer for what the GeoJSON mime type should be yet.)

nunziotocci:
It's working great for me!

klassenjs:
@theduckylittle Nice.

Issue #201, opened 2017-08-17
Catalog Layer Metadata link icon missing


mjreynolds:
If a catalog layer is configured properly in the mapbook via <metadata /> then when rendered a metadata icon link should appear in the geomoose catalog.  However, no icon is currently appearing.  The link is rendered though...  add the missing icon?

theduckylittle:
@mjreynolds  - Fix merged.

PR #202, opened 2017-08-17
Fix display of metadata icon.

klassenjs:
Looks like this got missed in f86ebea.
refs: #201


theduckylittle:
The build is failing on a change that is unaffected by this, I'm merging.

PR #203, opened 2017-08-21
Made catalog icons more touch friendly.

theduckylittle:
1. Made them bigger
2. Made the spacing wider
3. Put all the FA overrides into one declaration.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "geomoose",
-  "version": "3.0.0",
+  "version": "3.0.1-beta",
   "repository": "https://github.com/geomoose/gm3.git",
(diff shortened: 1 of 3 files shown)


klassenjs:
1) The Toggle Legend icon seems to have broken.

And some oddities I noticed and thought I should point out, but make no judgment about:

2) The toolbar icons moved down a bit.
3) The Favorite Stars don't change size.

theduckylittle:
1. rebase fixed the metadata icon (my bad)
2. I kept the icons that are on the "title row" the same size as the title font.  I thought making them larger than that would be odd looking.

klassenjs:
Legend icon still appears broken.

PR #204, opened 2017-08-21
Filters are now honoured for download features

theduckylittle:
This ensures that if a layer happens to have a particular
filter, then it can be downloaded.


klassenjs:
What's a test case for this?

theduckylittle:
I had to fabricate one locally and don't actually have a great example right now.

klassenjs:
Blindly merging then.  At some point a test for this would be nice (although I don't know how that would be implemented).

Issue #206, opened 2017-08-22
Tests Fail


klassenjs:
I suspect this is related to an OpenLayers update?

```
FAIL  tests/gm3/application.test.js
  ● Test suite failed to run
    TypeError: Cannot set property 'fillStyle' of null
      at node_modules/ol/imagetile.js:155:17
      at Object.<anonymous> (node_modules/ol/imagetile.js:158:2)
      at Object.<anonymous> (node_modules/ol/source/tileimage.js:2:18)
      at Object.<anonymous> (node_modules/ol/source/xyz.js:2:18)
      at Object.<anonymous> (src/gm3/components/layers/xyz.js:63:193)
      at Object.<anonymous> (src/gm3/components/map.jsx:1285:4061)
      at Object.<anonymous> (src/gm3/application.jsx:64:39)
      at Object.<anonymous> (tests/gm3/application.test.js:25:20)
      at handle (node_modules/worker-farm/lib/child/index.js:44:8)
      at process.<anonymous> (node_modules/worker-farm/lib/child/index.js:51:3)
      at emitTwo (events.js:125:13)
      at process.emit (events.js:213:7)
      at emit (internal/child_process.js:774:12)
      at _combinedTickCallback (internal/process/next_tick.js:141:11)
      at process._tickCallback (internal/process/next_tick.js:180:9)
```

PR #208, opened 2017-08-22
Fixing tests

theduckylittle:
1. Fixed travis to include the `canvas` module for testing.

   OpenLayers is using document.createElement('CANVAS') to create
   a dummy tile in the background. This causes issues within the
   test framework as jsdom does not implement the Canvas API fully
   without using the npm canvas module.

   That module is onerous to install on Windows so I have not made
   it a requirement in the package.json as this time.

2. Removed an unused inclusion of the Map component from Application
   which was causing that package to fail though it was not causing
   errors.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,37 @@
 #
-# (c) 2017 Dan "Ducky" Little
+# The MIT License (MIT)
+#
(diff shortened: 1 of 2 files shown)


theduckylittle:
Refs: #206

PR #209, opened 2017-08-25
Add reminder that WFS requires typename in map-source

tchaddad:
Not having this in a map-source that you want to be select-able will cause the WFS to throw an error.


theduckylittle:
Awesome, good catch!

Issue #214, opened 2017-09-01
zoomToExtent properties.boundedBy for point features zooms in too far


mjreynolds:
When a point feature is the basis for a Search, the resulting zoomToExtent button uses xmin ymin = xmax ymax - a point.  Clicking the zoom to extent button zooms the map way too close.

app.zoomToExtent([-10385229.40099363,5532031.259416136,-10385229.40099363,5532031.259416136])"

Is there a way to address this?  Can the templating calculate a buffer from properties.boundedBy or should the search service detect a point feature or xmin ymin = xmax ymax and render a more reasonable bounding box/zoom in?  Or could a minim extent be set for the whole application which the zoomToExtent wouldn't breach?

theduckylittle:
You could change the template to use `zoomTo` instead of `zoomToExtent`, `zoomTo` lets you specify the x,y, and resolution. So:

```javascript
app.zoomTo(geometry.coordinates[0],geometry.coordinates[1], 100);
```

mjreynolds:
I've used this:
app.zoomTo({{ geometry.coordinates.0 }}, {{ geometry.coordinates.1 }}, 100)"

Which renders this onclick value:
app.zoomTo(-10363951.34824544, 5596309.903175341, 100)"

When I click i get an error in the console:
Uncaught TypeError: t.slice is not a function at G.getState (view.js:757) ...

Somehow when this part is executing (view.js line 748):
/**
 * @return {olx.ViewState} View state.
 */
_ol_View_.prototype.getState = function() {
  var center = /** @type {ol.Coordinate} */ (this.getCenter());

center is not being set as an object with a slice() function.

See anything wrong with what I configured?

theduckylittle:
I found a bug in the Application class.  It's passing in `{lon, lat}` instead of `[lon, lat]` to `mapActions.move`.  Putting that on my to do list for today.

PR #218, opened 2017-09-06
Fix for broken zoom-to issue.

theduckylittle:
Previous version was passing in {lon, lat} when it should've
been [lon, lat]. This is likely a cruft that was never tested.

refs: #214


klassenjs:
Yay... More tests!

Issue #220, opened 2017-09-06
Could not hide the info-header/results-summary panel


theduckylittle:
Not every service needs to show the summary of layers/features.  That should've been toggle-able from the beginning.

PR #221, opened 2017-09-06
Added showHeader toggle to hide service summary

theduckylittle:
Previously, the summary would show for any service whether
or not it actually was querying layers + features. This was problematic
and is solved by allowing a service to set this.showHeader = false.

Refs: #220


klassenjs:
How does one set `showHeader = false`?

klassenjs:
Ok, figured it out...  How about an example in a service with the doxygen magic (even if just setting it explicitly to true)?

theduckylittle:
In the service class. This.showHeader = false.

Issue #178, opened 2017-06-28
Zoom To Extent drop down causing scrolling on FF (Linux)


klassenjs:
Firefox 45 (3px):
![ff45-g31229dc4](https://user-images.githubusercontent.com/165622/27641155-b2600c52-5be0-11e7-8b86-9aaaf7ea67b0.png)
![ff45-g31229dc4-down](https://user-images.githubusercontent.com/165622/27641157-b4a2f8a8-5be0-11e7-83e4-0c5b33c6965a.png)

Firefox 52 (9 px - also the "Search catalog" box got bigger under FF52)
![ff52-g31229dc4](https://user-images.githubusercontent.com/165622/27641179-c7198cd6-5be0-11e7-9a8a-1fe194a514aa.png)
![ff52-g31229dc4-down](https://user-images.githubusercontent.com/165622/27641183-c90b0cb8-5be0-11e7-950a-ba0b14b5a2c6.png)

klassenjs:
FF 54 on macOS does not show either the Zoom To causing scrolling or the oversized Search catalog box.

PR #223, opened 2017-09-24
CSS tweak to prevent the footer from overflowing.

theduckylittle:
This is an issue that makes itself readily visible in FF on linux.

### DIFF
--- a/examples/desktop/app.css
+++ b/examples/desktop/app.css
@@ -84,6 +84,7 @@ body {
     height: 20px;
     left: 0; width: 100%;
     border-top: solid 1px black;
+    overflow: hidden;
 }
 
 #version {


theduckylittle:
refs: #178

klassenjs:
Scrollbar is gone.  The drop down is clipped off a bit but still usable.  Good enough?
screenshot from 2017-09-24 18-11-30

theduckylittle:
We could do better but making the footer auto-height but it becomes a pretty deep rabbit hole with trying to prevent overflow into other elements of the UI.

theduckylittle:
so... yeah... this is better... if someone files a follow up looking for perfect / unusable we'll take another step forward. The real solution probably involves dropping the use of the select element as they are rendered VERY different across browsers and platforms.

klassenjs:
It is certainly the lesser of evils.

PR #224, opened 2017-10-02
New In-Memory Vector Layer Query Capabilities

theduckylittle:
1. Identify will now work on any vector or geojson type map-source. This includes the sketch layer and the newly included places.geojson file.
2. This PR adds and `auto="true"` template type which will sniff the features for their properties and spew out a template automatically.


theduckylittle:
refs: #227

klassenjs:
Auto appears to work as expected. 

A two small issues:
1) I can't find places.geojson.  Was this meant to be left in the mapbook?
2) I can't get identify to work when I upload a point geojson file (tolerance?)

klassenjs:
Tests pass locally.  Previous travis failure was due to yet another OL minor version bump that changed API.

Issue #225, opened 2017-10-03
GetLegendGraphic URL assembling can result in double ??


tchaddad:
I think this line adds a '?' to the base url without checking if there is already a '?' as part of the base:

https://github.com/geomoose/gm3/blob/e2324cd33e165036881c2d44234d1d8d14785fbc/src/gm3/components/layers/wms.js#L114

This can lead to a improperly formed URL which some servers will dislike:

http://oregongeology.org/arcgis/services/Public/SLIDO_3_2/MapServer/WMSServer??REQUEST=GetLegendGraphic&SERVICE=WMS&VERSION=1.1.1&WIDTH=250&LAYER=0&FORMAT=image%2Fpng

Same server without the double ??:

http://oregongeology.org/arcgis/services/Public/SLIDO_3_2/MapServer/WMSServer?REQUEST=GetLegendGraphic&SERVICE=WMS&VERSION=1.1.1&WIDTH=250&LAYER=0&FORMAT=image%2Fpng

klassenjs:
Note: a valid WMS service endpoint is considered an otherwise opaque URL
ending in either '?' or '&'.  So not including the final '?' (or '&') in a
mapbook is technically the spec violation.

That is to say I agree this is a valid bug.

theduckylittle:
PR for bug fix submitted.

theduckylittle:
Just to document it here, unless someone comes up with some more stuff. GeoMoose does a lot of URL manipulation internally. And while it is technically 'legal' to specify a URL with query parameters, GeoMoose expects those query parameters to be set in `<param>` tags in the `<map-source>`. This will normalize the parsing of a `?` in the URL but will give less predictable results if folks try to do stuff like `.../wms?QUERY=CUSTOM&KEY=012345`.

PR #226, opened 2017-10-03
Fix for extra "?" in URL.

theduckylittle:
refs: #225


klassenjs:
Summary of conversation between @theduckylittle and me:  This fixes a real world problem.  WMS URL handling is technically broken, but it was before too, effects no known real world server, so no regression.
Merging.

Issue #237, opened 2017-10-11
Bug with default values for the select input


theduckylittle:
When the default value was undefined then the first value of the select was considered the default. This is breaking because the onChange() is being called from render().

PR #238, opened 2017-10-11
fix for setting the default value when undefined

theduckylittle:
When defaultValue is undefined for a select, it was setting the
first value as the default but was during the render() cycle
which was causing an infinite loop.

refs: #237

### DIFF
--- a/src/services/select.js
+++ b/src/services/select.js
@@ -78,7 +78,7 @@ function SelectService(Application, options) {
         label: 'Query Layer',
         default: options.defaultLayer,
         filter: {
-            // do not require the layer be visible to select it.
+            // ensure that the layer is visible to prevent confusion.
             requireVisible: true,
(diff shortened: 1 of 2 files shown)


src/services/select.js
```diff
@@ -72,14 +72,14 @@ function SelectService(Application, options) {
...
 
     /** User input fields, select allows choosing a layer */
-    this.fields = [{
+    this.fields = options.fields || [{
         type: 'layers-list',
         name: 'layer',
         label: 'Query Layer',
         default: options.defaultLayer,
         filter: {
             // do not require the layer be visible to select it.
```
klassenjs: Not 100% but this looks like the comment doesn't match the code.

klassenjs: Github did that odd.  The comment on 81 vs next line.

Issue #242, opened 2017-10-11
Application.zoomTo should use "zoom" level and not resolution


theduckylittle:
General feedback is that the "zoom level" is a bit easier to digest than trying to make guesses at resolutions.

PR #243, opened 2017-10-11
zoomTo now zooms to a zoom level and not a resolution.

theduckylittle:
This fixes an expectation bug more than anything else.

After meeting with some users, and looking at the other
Javascript/GIS APIs, conceptualizing a "zoom" level is easier
than necessarily specifying a resolution.

refs: #242

### DIFF
--- a/src/gm3/actions/map.js
+++ b/src/gm3/actions/map.js
@@ -28,10 +28,11 @@
 
 import { MAP } from '../actionTypes';
 
-export function move(center, resolution) {
+export function move(center, zoom) {
     return {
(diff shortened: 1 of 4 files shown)


klassenjs:
This is a bit sketchy as it is both a bug fix (code doesn't match docs) and a change in how the exported API functions.  I'm merging because it doesn't look like it is used anywhere (yet). In the future, this should probably be handled by marking `move` deprecated with a note to use `move2` going forward sort of thing.  Then dropping `move` at the next major version.

Issue #235, opened 2017-10-11
Windows KML Uploads


theduckylittle:
Uploading the following KML works fine on Linux with Firefox and Chrome but fails on Windows with IE and Chrome.

[geomoose_1507742656590.kml.gz](https://github.com/geomoose/gm3/files/1376603/geomoose_1507742656590.kml.gz)

klassenjs:
That's a good trick.

klassenjs:
Not the core problem, but the KML file is missing a valid XML header.  `<?xml version="1.0"?>`

klassenjs:
Looks like it might be all KML files on Windows (tested W7 w/FF,Cr,IE).  A OGR produced one fails too.  File extension doesn't matter (tried txt and xml).

klassenjs:
My current thinking is it has to do with `if(file.type.indexOf('kml') ` failing to trigger on kml.

klassenjs:
file.type is `""` on Windows and `"application/vnd.google-earth.kml+xml"` on Linux.

klassenjs:
Not sure if should just detect with file.name.endsWith('.kml') or should sniff the contents. 
Try as JSON and if parsing fails, try again as KML?
Look for `{` or `<kml` or `<?xml` at the start of the file?

PR #248, opened 2017-10-25
Updated KML filename sniffer for windows compat

theduckylittle:
Windows was not checking the mime-type properly. File.type was
consistently set to ''. This uses the (inferior) filename test
in order to detect the type of parsers which should be used.

refs: #235

### DIFF
--- a/src/gm3/components/catalog/tools/upload.jsx
+++ b/src/gm3/components/catalog/tools/upload.jsx
@@ -159,7 +159,7 @@ class UploadModal extends Modal {
                     // input_format is defaulted to null and only
                     //  set if the file format can be inferred.
                     let input_format = null;
-                    if(file.type.indexOf('kml') >= 0) {
+                    if(String(file.name).split('.').pop() === 'kml') {
                         // sweet, KML file.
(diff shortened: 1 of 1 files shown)


klassenjs:
I nothing else, just for completeness of the record: I am ok with this approach, but any thoughts to sniffing the file contents rather than its name?

theduckylittle:
This test is a bit quickier-and-dirtier than sniffing the file. But there are some corner cases that I didn't want to start needing to chase down:

1. `<kml>` vs `<KML>` vs `<Kml>` in the tags. I'm not even sure if OL handles this properly but my understanding is that even `<kMl>` is theoretically valid. 
2. Searching for `{` first isn't totally valid either, though if it is the first character the likely hood the contents are JSON is pretty high.
3. There's a lot of room for failure of automatic detection with `indexOf` or a RegEx.

Filename sniffing seems like it should be reliable in 99% of cases. If a file is called `something.kml` it is pretty likely to be a KML file.

klassenjs:
Verified KML and GeoJSON uploads on W7: IE11, Cr61.  Linux: FF52.

Issue #256, opened 2017-10-25
Use HTML best practices for the index.html files


brentfraser:
Currently the character encoding for the mobile index.html is not set and displays a goofy character instead of "-" for zoom out.  So we could add a few things, for example:
```
    <!DOCTYPE html>
    <html lang="en">
    <head>
        <meta charset="utf-8"/>
```

brentfraser:
And an additional tag for mobile's index.html
`<meta name="viewport" content="width=device-width, initial-scale=1">`

brentfraser:
and actually, the viewport meta tag should go in the desktop index.html too so it will look good on all sizes of screens.

PR #263, opened 2017-10-25
Fixing up html-encoding issues.

theduckylittle:
As per #256

### DIFF
--- a/examples/desktop/debug.html
+++ b/examples/desktop/debug.html
@@ -1,6 +1,6 @@
 <!doctype html>
-<html>
-<head lang="en">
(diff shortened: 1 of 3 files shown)


klassenjs:
Isn't `lang` supposed to be on `<html>`?

https://www.w3.org/wiki/The_HTML_head_element

klassenjs:
One more reference: https://www.w3.org/International/questions/qa-html-language-declarations

theduckylittle:
Yeah, I just screwed up in some copy-and-paste zealotry.

Issue #258, opened 2017-10-25
Start Over Button


theduckylittle:
Need an easier way to clear that hash and refresh.

PR #264, opened 2017-10-25
Added a reload tool to the toolbar.

theduckylittle:
This isn't strictly necssary and I didn't think it warranted
creating a new service-js file.

This adds a reload button to the toolbar which will clear the "#" memory.

refs: #258


klassenjs:
I could get into a long debate if "Reload" and the circle icon are the best choice here.   I could also get into a debate about if it should ask "are you sure you want to wipe everything from your GeoMoose session and start with a clean slate?"

Thoughts anyone?  @elil usually has good insight into stuff like this.

But it does work.

elil:
I'm not well versed enough in the use of the hash to really have an opinion, or at least an informed one.  

"Reload" is perhaps vague to some people.  Quoting #258, "Start Over Button" is very clear but three words instead of one.  

Losing a bunch of accumulated work in one click of the button is very devastating.

theduckylittle:
I'm really find with calling this whatever. But it sounds like one of the biggest issues is that we should, likely, have a modal prompt asking "are you sure?".  "Start Over" may provide a clearer label. I'm also open to suggestions on an icon for it.

klassenjs:
I'd be happy with the modal.  

It does two things: 1. let's the user click it and gives them more explaination and 2. lets avoids consequences for one errant click.

The name/icon can be refined later if someone gets smart.

klassenjs:
Start Over or no change or other options are also fine with me for a name.

theduckylittle:
The updated code, as of this morning, is using "start over" as the label
and has the modal.

klassenjs:
Hate to nit-pick on this, but it is way more than reset Layers and Location to defaults.  If that were the case all the user would need to do to recover would be use the browser back button.

The bigger deal (and part of why it is useful) is this also resets everything else (e.g. drawings, queries, uploaded data, ...) none of which is recoverable.  The message should reflect this.

theduckylittle:
That's fine, please suggest some updated text.

klassenjs:
Well I already semi-jokingly did, but maybe:

Reset application to defaults?  This will reset selected layers and location to the application defaults, and remove any temporary drawings, queries, etc that you might have created.

klassenjs:
Or maybe something a little more common:

Are you sure you want to start over?  All unsaved work will be lost.

elil:
This sounds more common and understandable.  I'm trying to think about whether it should be phrased as a positive or negative but would need to find more examples.  So this is probably good.

Issue #251, opened 2017-10-25
Grid templates missing support for src


theduckylittle:
As the grid configuration can get otherwise complex it would be nice to have an option to configure the grid to work with auto templates, or at least add support for remote templates via the SRC attribute.

theduckylittle:
I don't think auto=true is necessarily practical (as there's no intelligent way to make sorting and filtering work). But PR #266 tweaks select to use remote templates with the grid.

PR #266, opened 2017-10-25
Reconfigured the grid to support remote templates.

theduckylittle:
1. The new default example uses a remote template for the grid.
2. Grid updated to detect when there is already a parsed object
   in the contents instead of a string.
3. select was changed to also request grid templates as available.
   This is a non-breaking backwards compatible change and the
   behaviour can be reverted by setting showGrid to false in the configuration
   section of select.

refs: #251


klassenjs:
This undos #265

theduckylittle:
I will merge it with master and repush.

theduckylittle:
fixed

klassenjs:
The interesting thing here is github didn't flag this as a merge conflict (before you resolved it).

Was git smart enough to do the merge (update the json file from the change that first happened in the mapbook) or did you need to fix this manually?

I probably wouldn't have cought the regression if I wasn't testing the PRs back to back.

theduckylittle:
I needed to do it manually. So, good catch!

Issue #254, opened 2017-10-25
Graceful WFS Failure


theduckylittle:
MapServer always returns a 200. Even for results which clearly have errors. We need to parse the Error and fail more gracefully.

PR #267, opened 2017-10-26
Added error handling for WFS requests.

theduckylittle:
MapServer returns WFS exceptions as HTTP status code 200,
this prevents the standard error detection from working. The code has
been updated to handle ows:Excpetion and wfs:Exception errors properly
and delivers them to the user.

refs: #254

### DIFF
--- a/src/gm3/actions/map.js
+++ b/src/gm3/actions/map.js
@@ -81,13 +81,14 @@ export function queryProgress(queryId) {
     };
 }
 
-export function resultsForQuery(queryId, layerName, failed, features) {
+export function resultsForQuery(queryId, layerName, failed, features, messageText='') {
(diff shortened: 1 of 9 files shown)


klassenjs:
I'm not seeing this working right now.  Not sure if it is the code or me right now.  My test was to change `ms:parcels` to `ms:parcelsii` in the mapbook and I don't see an error message or a console.log message.

theduckylittle:
Did you try it with *search* or with a *layer*? If you try it with a service than you'll get an error message in the tab. Layers will fail silently.

klassenjs:
Winner, I tried it with a layer not a search.  I presume catching the layer failure is "hard" because "OpenLayers"?

PR #272, opened 2017-11-10
added appropriate viewport settings

theduckylittle:
refs: #256

### DIFF
--- a/examples/desktop/debug.html
+++ b/examples/desktop/debug.html
@@ -2,6 +2,8 @@
 <html lang="en">
 <head>
     <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+
     <title>GeoMoose 3 Debug Instance</title>
 
     <link rel="stylesheet" type="text/css" href="../geomoose/node_modules/font-awesome/css/font-awesome.min.css" />
--- a/examples/desktop/debug.html
+++ b/examples/desktop/debug.html
@@ -73,9 +73,6 @@
     <div id="print-preview"></div>
 
     <script type="text/javascript" src="config.js"></script>
-    <!--
-    <script type="text/javascript" src="../geomoose/node_modules/openlayers/dist/ol.js"></script>
-    -->
     <script type="text/javascript" src="../geomoose/dist/geomoose.js"></script>
     <script type="text/javascript" src="../geomoose/src/services/identify.js"></script>
     <script type="text/javascript" src="../geomoose/src/services/search.js"></script>
--- a/examples/desktop/index.html
+++ b/examples/desktop/index.html
@@ -2,6 +2,7 @@
 <html lang="en">
 <head>
     <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1">
 
     <title>GeoMoose 3 Test Instance</title>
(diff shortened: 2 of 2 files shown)


theduckylittle:
As this is a small HTML tweak please ignore the travis failure. PR #273 will have a tweak for the travis conf that should allow it to run the tests successfully.

klassenjs:
I have to say, at least on my phone, this doesn't exactly improve the usability of the desktop version.  Without the meta line, at least there is enough screen room where the components fit.  It is small, but if I need to zoom to read or click something (or more often Chrome helps by auto-zooming part of the page you touch), I can.  With the meta line, things are more readably sized, but I can't zoom smaller to make things fit.

This is all portrait view, but landscape view is equally a mess.

Before:
screenshot_20171113-161136

After:
screenshot_20171113-161148
screenshot_20171113-161154

klassenjs:
This has been sitting in limbo for a while.  Is there any consensus?

Is this a net gain or net loss?

Is there a better way to implement it that doesn't make smaller screens unusable (vs just difficult)?

tchaddad:
I just checked the demo from both Firefox and Chrome on a small phone (iPhone SE), and I'm not seeing the same problem, but maybe I'm missing the point?

Firefox for iOS
image

Chrome for iOS
image

tchaddad:
And to be more clear - I'm not seeing any wrapping when I zoom in on the UI in either browser (just goes off screen). I suppose more optimal would be if on a small screen the toolbar div wrapped around and pushed the catalog and map down, so that all tools could be seen?

klassenjs:
The pull request hasn't been merged so it isn't on demo.geomoose.org.  You'd have to make the change to a local copy of GeoMoose to see it.

klassenjs:
I'll make a temporary copy with the change on demo.geomoose.org to help facilitate discussion.

klassenjs:
https://demo.geomoose.org/viewport/desktop/

elil:
Checking on firefox 47.0 on Android (pretty dated), https://demo.geomoose.org/viewport/desktop/ is worse than https://demo.geomoose.org/master/desktop/ because the top menu wraps inaccessibly behind the tabs just below: Catalog, Favorites, Visible, and Super Tab.  

Viewport url with menu wrap behind things: 
screenshot_2018-03-08-13-34-58



Demo currently (does not wrap and can access tools):
screenshot_2018-03-08-13-35-34

Demo currently (can zoom in and scroll to different areas of screen):
screenshot_2018-03-08-13-35-44

elil:
Same on iOS 10.3.3 on Safari

tchaddad:
Yep can confirm the same here - I guess the small screen behavior prior to this PR is better than after for smallest screen. Don't have a medium tablet handy to test with at the moment...

klassenjs:
More data points:

On my Nexus 7:  This PR makes things ever so slightly bigger in landscape and it remains usable.  In portrait it does the same wrapping thing from above, while the before is much smaller, but usable.

As far as I can tell, this PR makes no difference on my 10" Win 10 tablet under Firefox, Chrome, and Edge.

The wrapping occurs on the desktop too with a window width < 840px, but can be resolved by zooming out (w/```Ctrl -```).

klassenjs:
Closing this as it appears to cause more problems than it fixes at the moment.

PR #273, opened 2017-11-10
Added a control to allow clearing selections

theduckylittle:
After all queries have been removed from the selection tab,
the selection polygon would continue to be displayed.

This new functionality fixes that issue by adding a button
allowing the user to clear the polygon.  This was done as a bit
of a compromise. Ostensibly, a complex selection area may need to be
re-used even if the user wants to clear teh current results.  Automatically
removing the selection area would've made that impossible.

Steps to use:
1. run a query with a drawn selection feature.
2. click "delete" / trash-can icon on the results.
3. See that there is now a polygon on display.
4. click the 'clear' button in the service manager tab.

refs: #257

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,5 +33,5 @@ node_js:
 before_script:
     - sudo apt-get -qq update
     - sudo apt-get install -y libcairo2-dev libjpeg8-dev libpango1.0-dev libgif-dev g++
-    - npm install canvas
+    - npm install canvas@1.6.7
     - grunt lint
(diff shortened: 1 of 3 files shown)


elil:
This seems like fine grained control!  Probably appreciated by those who know what it is and use it, probably ignored by many.  Out of curiosity, what happens when you use this order for the steps?


1. run a query with a drawn selection feature.
2. click the 'clear' button in the service manager tab.
3. See that there is now a polygon on display.
4. click "delete" / trash-can icon on the results.

theduckylittle:
The clear button is disabled when there is no polygon available to clear. If you do hit the clear button while drawing, it will just remove the currently in progress selection.

elil:
👍

PR #274, opened 2017-11-27
Allow any bing layer name

brentfraser:
Add the ability to specify other Bing layers.  Included examples in the mapbook.xml/

### DIFF
--- a/examples/desktop/mapbook.xml
+++ b/examples/desktop/mapbook.xml
@@ -381,6 +381,9 @@
     </map-source>
 
     <map-source name="bing" type="bing">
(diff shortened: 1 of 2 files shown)


theduckylittle:
thanks a bunch @brentfraser !

brentfraser:
Happy to do it, but an issue with my looping thru the layers when several are turned on.  It really only requests the "last" one specified in the MapSource.  Not a huge problem since a user will see only one base layer no matter how many are turned on (it just might not be the one they expect).  Could be solved by implementing radio buttons for the base layers.

PR #280, opened 2018-01-25
Fix spelling errors in the docs.

klassenjs:


elil:
Nice work!  Did you do this by machine or eye?  We could look at some machine options as well.  GDAL did a big [clean up](https://trac.osgeo.org/gdal/ticket/6230) at one point.

klassenjs:
A little of both.  I noticed a couple of errors at first and then decided to run everything through aspell.

theduckylittle:
I approve.

Issue #276, opened 2017-12-10
404 for site.css


tchaddad:
Since we don't provide this file, maybe we should comment it out in the index.html file?

The documentation explains what site.css is supposed to be for, and explains how to add it, so referencing it in the index file as if it already exists, seems contradictory to the docs...

elil:
(Now we're getting offtopic: Pertaining to online search and other Sphinx niceties, there are Jekyll plugins for things like that but the plugins aren't permitted to run on GH Pages.  We could build locally and push but I don't think that is worth it.)

PR #283, opened 2018-01-31
Drop the site.css reference.

theduckylittle:
It created an erroneous 404 and the docs suggest adding it.

refs: #276

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,8 +30,10 @@ node_js:
 # setup some solid deps.
 
 # ensure the code lints clean.
-before_script:
+before_install:
     - sudo apt-get -qq update
     - sudo apt-get install -y libcairo2-dev libjpeg8-dev libpango1.0-dev libgif-dev g++
+
+before_script:
     - npm install canvas@1.6.7
     - grunt lint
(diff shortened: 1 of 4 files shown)


klassenjs:
OK... This is a really messy PR.

Drop site.css reference: I'm OK with this.  I might suggest commenting it out instead of deleting it, but not a big deal.

Drop pre-pub and change to https instead of git+https:
1) Dropping prepublish breaks the build of the npm package (which in turn breaks the build of the ZIP and tgz releases).  I know prepublish is deprecated, but `npm pack` needs to build.  It looks like it should be changed to a `prepare` (equivalent) or `prepack` (doesn't run automatically on npm install) script for node 6+.  (For the record, it is using prepublish because older node versions didn't support prepare and they were still a supported build target at the time.)  I've been meaning to get to this anyway.

2) git+https -> https may work, but it is contrary to the npm docs for requesting a git repo (which say to use git+https or the github shortcut)  Although it does appear to still fetch the repo.   https://docs.npmjs.com/files/package.json#git-urls-as-dependencies

3) Canvas was added as a dependency which breaks the Windows build

Fix Travis:  I can't tell that this actually fixed travis because travis still failed.

I'm kinda thinking we should mulligan this PR probably as a separate one for site.css and one for Build Fixes.  So, I'd say lets discuss a little to pick a direction (prepare vs prepack, and git+https vs https vs user/repo) and then either you or I can make the new PRs.

klassenjs:
And I presume we want to leave the canvas dependency as a travis before_script item (although something more windows friendly would be nice long term).

klassenjs:
Closing.  Replaced with #285 and #286.

PR #285, opened 2018-03-03
Attempt to fix travis build errors.

klassenjs:
Looks like implicit `npm install` is no bueno.


klassenjs:
Note: I have played with the order of things.  I know that `npm install` happens automatically early on, but `npm install canvas` is definitely causing some chaos if not followed by another `npm install`.    Note the logs mention one package removed after installing canvas.

klassenjs:
Also, stops annoying @theduckylittle with `npm install` automatically calling `grunt build`.

Issue #290, opened 2018-03-26
Linter/CSS


klassenjs:
Linter doesn't catch spaces vs. tabs in examples/desktop/tabs.css

PR #291, opened 2018-03-28
Added a tab check to the less linter.

theduckylittle:
Looks for \t's throws errors when they are there.

refs: #290


klassenjs:
Not at the computer right now to check this... Is anything checking /examples/**

theduckylittle:
Not yet. But that will likely come when I re-do the examples.  Since we need to jettison the current mobile demo.

klassenjs:
Ok.  Merging.

I'd leave #290 open as a reminder to lint examples.

theduckylittle:
Actually, I think it might be better to setup a "revamp examples" project that features a lint task.

klassenjs:
Fair enough.  Set it up however you think it is most useful.  I just don't want it forgotten.

Issue #287, opened 2018-03-08
Select results grid does not work following identify


elil:
Open demo
Identify a parcel
Select parcel(s)
Hit "Go"
Get Super Tab that says, "Parcel selection results are shown in the results grid. "
There is no results grid, expect results grid at bottom of page.

klassenjs:
Confused:  If I use the Identify tool, the results show up in the Super Tab as soon as I click on a parcel.  (Note: if I click "Go" without clicking on a parcel some strange stuff happens that probably deserves it's own issue).

If I use the Select Features tool to select a parcel, I see the message you indicated in the Super Tab, but also get a results grid at the bottom of the page.

If I use the Select Features tool and create a selection that doesn't select any parcels, it does what you said (and also says 0 features were returned in the Super Tab).

klassenjs:
Ok, Got it.

1) Click Identify Tool
2) Click on a parcel (Super Tab shows results of identify as expected)
3) Click Select Features Tool
4) Draw selection of a parcel(s) on map, and click "Go"
5) Results in N features selected in Super Tab, Selected Parcels highlighted on map, but no results grid.

tchaddad:
Eli's combo only occurs if Select is not the first tool run. 

I ran into it as well, but was doubly confused because in other instances, the results would be provided in the table, but the table would just be minimized, and I didn't realize it:

![results_collapsed](https://user-images.githubusercontent.com/1042618/37178460-6c1e3a84-22d7-11e8-8010-72bb520757ac.PNG)

Could the problem be similar in this case, except with the collapse icon not even drawing to indicate the table is there?

klassenjs:
Note: The identify creates an error in the console about:  ```l.templates.gridColumns is undefined```

theduckylittle:
I believe this is fixed as fall-out from fixing identify:
- https://github.com/geomoose/gm3/pull/292

klassenjs:
#292 fixed this for me.  @elil?

elil:
Yes, appears fixed.  Thanks.

Issue #288, opened 2018-03-08
Identify with no selection crashes


klassenjs:
Steps:
  1) Reload/Load GeoMoose fresh (to make sure there is no cached selection).
  2) Click the Identify Tool
  3) Click "Go" in the Super Tab (do not click on the map first).

Results: The Super Tab sits and spins, the map disappears, a couple of variable is undefined errors in the console. 
  
Expect: It's an undefined operation so I don't know what it should do instead, but not crash would be good.

PR #292, opened 2018-03-28
Fix invalid geometry for identify.

theduckylittle:
1. Checks to ensure a geometry exists and that it is valid for identify.
2. Creates a better failure notificaiton.
3. Add undefined check for the grid-columns.

refs: #288


klassenjs:
Tested fixed: #287 and #288

PR #301, opened 2018-04-09
Add lock file to prevent mis-versioned builds

theduckylittle:
Too many deep deps getting wonky. This has become, unfortunately,
necessary.


theduckylittle:
PR #300 should get merged first as that commit is also included here for consistency sake.

Issue #318, opened 2018-09-21
Precision: 0 is the same as not specifying precision, not 0 decimal places


klassenjs:
```
    var point_projections = [
        {
            label: 'X,Y',
            ref: 'EPSG:200068',
            precision: 0
        },
```

PR #319, opened 2018-09-25
Fix for precision handling

theduckylittle:
- Centralizes format output logic.
- Fixes issue with precision test.

refs: #318


klassenjs:
Hey, negative precisions even work.  Cool!

PR #322, opened 2018-09-25
Adding selenium tests for browser stack.

theduckylittle:
This is meant to be a _very basic_ smoke test
to ensure that IE will load the application without
error.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,10 @@
 # is any build artifacts that are needed at runtime should be included
 # in npm but excluded from git.
 
+# Python n things
+env/
+.env
+*.pyc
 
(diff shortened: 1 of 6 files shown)


klassenjs:
Not sure it matters, but both the python runner scripts have a bug where they fail to driver.quit() if there is an exception.

klassenjs:
@theduckylittle merge if happy with my commits.

klassenjs:
Only general comment is if this could be written in node.js, that might be better instead of adding Python as a dep.  (Although I see https://www.npmjs.com/package/selenium-webdriver needs local binary builds and a .jar which seems worse).

In any case, I see this as far better than nothing, and a possible switch to JS is not a huge deal.

PR #334, opened 2018-10-15
WIP: Add USNG Search extension

klassenjs:
Work in Progress. Do not merge.  

Making this a PR for discussion tracking.

The idea is given an optional starting point (only used in the case of a truncated coordinate) and a USNG coordinate to return the USNG grid "square" (in the results and on map) that represents that coordinate.

### DIFF
--- a/examples/desktop/app.js
+++ b/examples/desktop/app.js
@@ -107,6 +107,7 @@ app.loadMapbook({url: 'mapbook.xml'}).then(function() {
     // there is also a BingGeocoder service, but requires
     // signing up for Bing and getting an appropriate usage key.
     app.registerService('geocode', OSMGeocoder, {});
+    app.registerService('usngsearch', USNGSearch, {});
     app.registerAction('findme', FindMeAction);
 
     app.registerAction('fullextent', ZoomToAction, {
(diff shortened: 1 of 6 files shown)


klassenjs:
Some test points:
15TVK919779 should return a 100m box over a dome.
15TVK91937797 should return a 10m box over the peak of the dome.

If a point is selected within 50km of the dome the 15TVK part can be left off.

klassenjs:
15TVK8326 is a good multi-parcel spot in the demo.

klassenjs:
Update, the two big issues at the moment are:
1) Needing to modify toolbar.less to make the toolbar icon work.  This isn't an issue if we bundle this with gm3 but is a bit on an issue for external services.

2) Having to include (a hacked version of) USNG2 separately instead of using the USNG3 library that is already present.  USNG2 was used because USNG3 assumes a build step as it assumes it will be included via 'require' (which is probably an issue with USNG3) and services are not "built".  I will also need to add an exported function to USNG3, but that's a separate issue.

klassenjs:
This should be evaluated as a plugin now that #465 is merged.

Issue #323, opened 2018-09-28
Identify with no identifiable layers fails


klassenjs:
If there are no visible layers with an identify template, the identify tool just spins with the loading arrows.

christian4423:
I am getting this also. Did you check your request and view the response?

PR #335, opened 2018-10-16
Fix for Identify

theduckylittle:
- Giant pile of React/Redux fixes
- Fixes the broken identify.

Refs: #323


klassenjs:
This may be a different case, but I am still seeing #323 like behavior with:

1) Open GeoMoose
1) Turn off all layers on visible tab
1) Identify, click point

theduckylittle:
That was the exact behaviour I was testing with...

klassenjs:
Miss an add/push?

klassenjs:
I'm on dbc956b

klassenjs:
Tested with:
```
rm -rf dist coverage node_modules
npm install && grunt build
```

PR #340, opened 2018-11-12
search field values validation

cabesuon:
This pull request allows the user to add validation to the search fields. It is achieve using the new `validateFieldValues` method of the search service. This method takes as an argument the field values (object `{fieldName: value}`) and returns a boolean value that indicate if the input is valid and a correspondent error message (object `{valid: boolean, message: string}`).

The main changes are in `serviceForm.js`:
- two new state property were added `validateFieldValuesResultMessage` and `lastService`,
- a div message for the error was added, that shows in case `validateFieldValuesResultMessage` is not null
- an invocation to the service `validateFieldValues` is made, if exist, on the click event of the go button
- the component `UNSAFE_componentWillUpdate` method was added to udate the state of the new properties if it is a new service


theduckylittle:
Sorry for the delay on this one! Looks great! Merging in...

PR #344, opened 2018-12-20
OpenLayers 5 Upgrade, Tests no longer require canvas

theduckylittle:
1. Upgrade to OpenLayers 5.
2. Map tests are skipped when `node-canvas` is not present.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,12 @@ grunt build
 npm test
 ```
 
+*Tests will skip the map unless canvas is installed. To perform a more complete test ensure node-canvas 1.x is installed*
+
+```
+npm install canvas@~1.0.0
+```
+
 ## License
 
 GeoMoose3 uses an MIT-style licence. You can read the contents of it [here](./LICENSE)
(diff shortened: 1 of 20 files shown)


klassenjs:
I am seeing test failures with canvas installed.  I haven't dug into this yet.

```
Error: Not implemented: HTMLCanvasElement.prototype.getContext (without installing the canvas npm package)
```

klassenjs:
Not sure if I am not seeing the same version of things, but Select Features is failing, and measure is returning an empty tab instead of measurements.

klassenjs:
Search parcel is failing too..... The only think I see that I can make sense of is `this.mapDiv is null`

theduckylittle:
You shouldn't be seeing that error on the latest branch... the getContext is polyfilled... on a clean directory I don't have a problem.

klassenjs:
I cleaned all the stuff in gm3, unless it is in `~/.node-hides-stuff-in-silly-spots`.

klassenjs:
Things look good with the latest push.

elil:
Wow, nice work! :+1:

PR #345, opened 2019-01-16
Adding Min/Max-resolution support to catalog + map

theduckylittle:
1. Added suppot for `minresolution` and `maxresolution`
   to map-sources and catalogs.
2. Added default CSS to style the layers.
3. Added example to mapbook.


klassenjs:
How does this play with search/select?

klassenjs:
Looks like Identify will Identify layers that are out of range.

PR #350, opened 2019-02-01
Added configurability to Selection and Results

theduckylittle:
* Added the code to allow configuring them from the application
   constructor.
 * Moved the style defintions to a defaults module for easier
   reference, maintenance, and testing down the road.
 * Added a doc to refer people to.

### DIFF
--- a/docs/howto/change-highlight-colors.md
+++ b/docs/howto/change-highlight-colors.md
@@ -1,0 +1,93 @@
+# Changing the selection and highlight colors
+
+## Differences between GeoMoose 2.X and 3.X
+
+In GeoMoose 2.X there was a layer called `highlight` that was defined in the mapbook,
+occasionally users would remove this layer and services would cease to function.
+To prevent that from happening GeoMoose 3+ will automatically define a layer
+named `results` and `results-hot`. The `results-hot` layer is used when the user
+has selected a subset of results.
+
+## GeoMoose vector styles
+
+All vector styles are styled using the Mapbox GL Styles format. It is Javascript/JSON
+friendly and well documented here. TODO: INSERT LINK.
+
+## Change the highlight color
+
+Pass `resultsStyle.highlight` into the `loadMapbook` options.
+
+Example:
+
+```javascript
+
+// use a pretty blue color instead of the default yellow
+const highlightColor = '#008080';
+const highlightStyle = {
+    'circle-color': highlightColor,
+    'circle-stroke-color': highlightColor,
+    'line-color': highlightColor,
+    'fill-color': highlightColor
+};
+
+var app = new gm3.Application({
+    mapserver_url: CONFIG.mapserver_url,
+    mapfile_root: CONFIG.mapfile_root,
+    resultsStyle: {
+        highlight: highlightStyle,
+    }
+});
+
+```
+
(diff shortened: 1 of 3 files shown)


docs/howto/change-highlight-colors.md
```diff
@@ -0,0 +1,93 @@
...
+has selected a subset of results.
+
+## GeoMoose vector styles
+
+All vector styles are styled using the Mapbox GL Styles format. It is Javascript/JSON
+friendly and well documented here. TODO: INSERT LINK.
+
+## Change the highlight color
+
+Pass `resultsStyle.highlight` into the `loadMapbook` options.
```
klassenjs: This doesn't match the example code.

klassenjs: gm3.Application != loadMapbook

src/gm3/application.js
```diff
@@ -248,15 +229,15 @@ class Application {
                 url: options.url,
                 type: 'xml',
                 success: (response) => {
-                    this.populateMapbook(response);
+                    this.populateMapbook(response, options);
```
klassenjs: populateMapbook only takes one argument

src/gm3/application.js
```diff
@@ -248,15 +229,15 @@ class Application {
...
                 }
             });
 
         } else if(options.content) {
             // this is just straight mapbook xml
             //  TODO: Maybe it needs parsed?!?
             const populate_promise = new Promise((resolve, reject) => {
-                this.populateMapbook(options.content);
+                this.populateMapbook(options.content, options);
```
klassenjs: populateMapbook only takes one argument

docs/howto/change-highlight-colors.md
```diff
@@ -0,0 +1,93 @@
...
+In GeoMoose 2.X there was a layer called `highlight` that was defined in the mapbook,
+occasionally users would remove this layer and services would cease to function.
+To prevent that from happening GeoMoose 3+ will automatically define a layer
+named `results` and `results-hot`. The `results-hot` layer is used when the user
+has selected a subset of results.
+
+## GeoMoose vector styles
+
+All vector styles are styled using the Mapbox GL Styles format. It is Javascript/JSON
+friendly and well documented here. TODO: INSERT LINK.
```
klassenjs: TODO

klassenjs: Wow, that's a hard link to find.  MapBox GL Styles are well documented, but I gather we must be using some subset of MapBox GL Styles since overall it looks like it more or less is a JSON version of a GeoMoose 2 mapbook.

klassenjs: https://docs.mapbox.com/mapbox-gl-js/style-spec/#layers

If I had to guess it would be some combination of 'paint' and 'layout' properties (although I don't think we are splitting them out into separate categories).

klassenjs: Fair enough.  The other thing is if updating it breaks the current styles, it would force us into major release which is somewhat undesirable.

klassenjs:
Did we intend to change the colors in the demo (app.js)?  I'd be tempted to leave the lines in app.js but comment them out.  Otherwise, it looks ok to me.

Issue #315, opened 2018-08-26
Adding OR connector to search


cabesuon:
This is a question and suggestion at the same time. I don't know if this is already solved, so excuse me it it is. I am sure there are better ways to solve it, this is just an example.

Regards
Manuel

User Case Example: The user wants to search for one or several parcels of an administrative limit. A parcel is search by a combination of a administrative limit name (Departamento) and a parcel number (Padrón).

Input Example:
Departamento: COLONIA
Padrón: 11938 7917 11977

Right now I could not find a way of doing this. So I have to change a bit the code of map.jsx, and aloud user to enter several numbers separated with a space character on the number input.

-> app.js - prepareFields function of parcel search *

        prepareFields: function(fields) {
            var query = [];
            // departamento
            if(fields[0].value !== '' && fields[0].value !== undefined) {
                query.push({
                    comparitor: 'eq',
                    name: fields[0].name,
                    value: fields[0].value
                });
            }
            // padron
            if(fields[1].value !== '' && fields[1].value !== undefined) {
                let parcels = fields[1].value.trim().split(' ');
                let parcel_filter = null;
                if(parcels.length > 1) {
                    parcel_filter = {
                        connector: 'or',
                        rightCondition: {
                            comparitor: 'eq',
                            name: fields[1].name,
                            value: parcels[0]
                        },
                        leftCondition: {
                            comparitor: 'eq',
                            name: fields[1].name,
                            value: parcels[1]
                        }
                    }
                    for(let i = 2, ii = parcels.length; i < ii; i++) {
                        parcel_filter = parcels_filter = {
                            connector: 'or',
                            rightCondition: parcel_filter,
                            leftCondition: {
                                comparitor: 'eq',
                                name: fields[1].name,
                                value: parcels[i]
                            }
                        }
                    }
                } else {
                    parcel_filter = {
                        comparitor: 'eq',
                        name: fields[1].name,
                        value: fields[1].value
                    };
                }
                query.push(parcel_filter);
            }
            return query;
        }

-> map.jsx - function wfsGetFeatureQuery - added and commented code

        // to support binary connectors 'and' and/or 'or'
        let resolveFilter = function(filter) {
            if (filter.hasOwnProperty('comparitor')) {
                return filter_mapping[filter.comparitor](filter.name, filter.value);
            } else if (filter.hasOwnProperty('connector')) {
                if (filter.connector === 'and') {
                    return ol_filters.and(
                        resolveFilter(filter.rightCondition),
[...]

PR #353, opened 2019-02-05
Added support for OR operator in queries (and travis fixes)

theduckylittle:
Refactored WFS filter building code to support
OR-type queries.

Even with tests!


theduckylittle:
Refs: https://github.com/geomoose/gm3/issues/315

klassenjs:
Note: `xmlshim` is native code (so probably breaks building on Windows).

chughes-lincoln:
Some docs for how to configure this would be very helpful. I'm happy to make them but I haven't yet figured out how to get it working on my end. Is the end result supposed to be allowing OR support within a field? (i.e. if you have a parcel id/PIN field, you could search for multiple parcel id's separated by some delimiter?)

PR #355, opened 2019-02-05
Prevent zoom-to-results from going too far

theduckylittle:
This can be a praticular problem when returning a single
point (as in from a search) and zooming to the results will
take the user to a view which is not useful.  Adding a minimum
bounding box size prevents that from happening.

### DIFF
--- a/src/gm3/components/serviceManager.js
+++ b/src/gm3/components/serviceManager.js
@@ -64,7 +64,7 @@ function normalizeSelection(selectionFeatures) {
 /** Get the extent of a query's results.
  *  All features must have a boundedBy property.
  */
-function getExtentForQuery(results) {
+function getExtentForQuery(results, minSize = 150) {
     let extent = null;
 
     for(const path in results) {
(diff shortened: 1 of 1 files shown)


klassenjs:
Can a user change the default?  Per layer?

theduckylittle:
Can we do that in a follow up? I feel as if we need to add a "config" reducer that allows changing these defaults better but there aren't a lot of 'feed points' to that part of the code.

src/gm3/components/serviceManager.js
```diff
@@ -82,6 +82,18 @@ function getExtentForQuery(results) {
             }
         }
     }
+
+    if (extent[3] - extent[1] < minSize || extent[2] - extent[0] < minSize) {
```
klassenjs: This should probably be separate conditions for x and y.  If the real extent is 1 x 1000 (say a street segment) this will shrink it to 150x150.  (Or... maybe it should be an `&&` instead of a `||`)

klassenjs:
I was looking at the code and wondering if there was a path in.  I suppose we can make that a separate issue... based on geomoose-users, I expect configuring it will come up quickly.

Issue #357, opened 2019-02-05
Select pipelines gives an error


elil:
1. Catalog --> turn on pipelines
2. Select features --> Draw polygon --> [draw generous polygon] --> Query Layer --> Pipelines --> Go
3. No valid selection geometry.

theduckylittle:
Doh! This was a simple one. The Pipelines shouldn't be there. The layers-list drop down was not honouring the "require select template" filter.

elil:
Looks good in the demo now.

elil:
It looks like my review and closing based on the demo was mistaken.  I failed to make the pipelines visible before testing, that is why they were not present in the Query Layer.  (If only I had followed my own detailed steps above!)

Reopening to let someone else test properly.

klassenjs:
Try it now.  I think you got ahead of the buildbot.

elil:
Thanks, looks good now!  Now, I'm closing...

PR #359, opened 2019-02-06
Restore layerslist input filter

theduckylittle:
The layerslist input filter was dropped at some point,
this fixes that oversight.

refs: #357


klassenjs:
I am getting console.log errors (and general unpleasantness):

1) Turn on Pipelines
2) Click `Select Features` tool
3) Click `Select From:` radio button.

Issue #361, opened 2019-02-06
Scalebars seem Incorrect


tchaddad:
I might be crazy, but the scalebars in the demo seem to be shorter than then should be by about 25%.

This can be confirmed by measuring against a feature of known length, or even using the measure tool, which by comparison to known data, is pretty accurate

theduckylittle:
I can verify the behaviour on master.

theduckylittle:
Hm... looks like GeoMoose is passing the correct values to MapServer...

klassenjs:
It doesn't happen to be wrong by cos(latitude) does it?  (.707 at 45°)

tchaddad:
I wondered the same thing. 

Or possibly that what is being passed is not being passed with enough decimal values (if the math is being done in 4326).

theduckylittle:
I think it might be a technique issue. GeoMoose is doing projection to valid measurement projections whereas I believe MapServer may be actually just blindly using EPSG:3857.

theduckylittle:
I'm leaving this open despite the PR...

tchaddad:
Relevant thread: https://lists.osgeo.org/pipermail/mapserver-users/2018-January/080474.html

tchaddad:
Would using OpenLayers to add the scalebar be a future enhancement option?

https://openlayers.org/en/latest/examples/scale-line.html

klassenjs:
I suspect that MapServer is treating the X,Y of 3857 as literal meters when it generates the scalebar.  IIRC the scalebar is controlled by the `MAP->UNITS` setting [1] and not by the actual projection definition.  This is more-or-less correct in reasonably behaved local projections (and is correct in an equal area projection), but is a bad assumption in a Mercator projection.

[1] From https://mapserver.org/mapfile/map.html
> UNITS [dd|feet|inches|kilometers|meters|miles|nauticalmiles]
>    Units of the map coordinates. Used for scalebar and scale computations. Nauticalmiles was added in MapServer 5.6.

theduckylittle:
@tchaddad - We avoided using the scaleline control in the past because it doesn't print well... but it's definitely worth researching.

klassenjs:
Is that still an issue with GM3?

theduckylittle:
Yes it is. The scaleline is still rendered as DOM and not as a graphic on the map. So when we scrape the map-canvas for a picture we would still need to render the scaleline our selves.

tchaddad:
[...]

PR #362, opened 2019-02-06
Document issue with scalebars and measurements.

theduckylittle:
Doc to cover issues we've found with EPSG:3875.

refs: #361

### DIFF
--- a/docs/projections-and-scalebars.md
+++ b/docs/projections-and-scalebars.md
@@ -1,0 +1,28 @@
+# Projections, Measurement, and Scalebars
+
+GeoMoose uses the EPSG:3857 projection for displaying maps.
+This choice was made to ensure that users could use OpenStreetMap,
+Bing, and other popular basemap layers. This is the same projection
+that is used by other popular web mapping applications.
+
+Problematically, this Mercator projection can show a number of artifacts
+in daily use.
+
+## Measurements
+
+Measuring between two points on the screen will not necessarily yield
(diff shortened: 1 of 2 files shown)


docs/projections-and-scalebars.md
```diff
@@ -0,0 +1,28 @@
...
+Bing, and other popular basemap layers. This is the same projection
+that is used by other popular web mapping applications.
+
+Problematically, this Mercator projection can show a number of artifacts
+in daily use.
+
+## Measurements
+
+Measuring between two points on the screen will not necessarily yield
+an accurate measurement.  As the map goes north to south measuring points
```
klassenjs: north to south -> further from the equator

docs/projections-and-scalebars.md
```diff
@@ -0,0 +1,28 @@
...
+will yield decreasing accuracy.  GeoMoose accounts for this fact and will
+perform the necessary re-projection in order to provide the user with
+meaningful numbers.
+
+## Scalebars
+
+The scalebars included with the demo are rendered by MapServer.  MapServer
```
klassenjs: I'm not sure about my square pixel argument (although I don't see yet why that would be different from 4326).   I think it particular issue is more related to mapserver assuming the X,Y of the map coordinates are in whatever the UNITS is set to (METERS for us) rather than running it through proj4 like it does for the map data.

theduckylittle:
Closing this as I think that #371 fixes it better than docing it.

Issue #366, opened 2019-03-01
Buffers seem too small


tchaddad:
Following up from the discussion in https://github.com/geomoose/gm3/issues/361, buffer distances seem too small by 25-30%, which seems to be the same error around scalebars.

leaving this as a stub, and will come back and attach screenshots in the next few days.

tchaddad:
Here's an example:

- Browse the demo for some adjacent parcels that look the same width
- measure to get approximate widths to confirm (in this case it was ~1300 feet)
- set the buffer distance to 1300 feet and choose to select with a point
- click in a corner between two of these similarly sized polygons
- buffer draws, and clearly is too short to reach the expected distance of ~1300 to the edge of the neighboring polygon
- distance seems off by about the same % as mentioned in the scalebar issue (https://github.com/geomoose/gm3/issues/361)

Like so:

![image](https://user-images.githubusercontent.com/1042618/53620583-58b03c00-3ba8-11e9-9d22-ecaef68267a6.png)

theduckylittle:
The aggravating part is that the measurement tools and the buffers are likely _more accurate_ than what is being rendered on the map.  This is pretty high on my 'to investigate' list but I just haven't had an open window of opportunity yet.

chughes-lincoln:
Looks like this is fixed in the current demo

PR #371, opened 2019-04-15
Fixing buffer and scale distances

theduckylittle:
1. Fixed the buffer calculation by doing the manual
   conversions in/out of traverse mercator.
2. Added the ability to enable openlayers scaleline
   indicator which was also correctly 'measuring' the
   distance represented on the map.
2b. I added a config reducer as a catch-all of app settings that
    may not necessarily be best interpreted by the mapbook.
    This may be fluid and should likely allow the user to change
    settings if a developer gets ambitious.
3. Dropped the misleading mapserver scalelines from the
   desktop demo mapbook.

### DIFF
--- a/examples/desktop/app.js
+++ b/examples/desktop/app.js
@@ -31,7 +31,13 @@
 
 var app = new gm3.Application({
     mapserver_url: CONFIG.mapserver_url,
-    mapfile_root: CONFIG.mapfile_root
+    mapfile_root: CONFIG.mapfile_root,
+    map: {
+        scaleLine: {
+            enabled: true,
+            units: 'imperial'
+        }
+    }
 });
 
 app.uiUpdate = function(ui) {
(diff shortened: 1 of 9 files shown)


theduckylittle:
@tchaddad - For some reason I can't make you a reviewer but I'd like you to see this!!!

tchaddad:
Nice - will check it out!

elil:
Do we still do refs: #361 and #366?  (That might give someone on the issue a ping.)

theduckylittle:
Thanks for catching the refs @elil !!! Those are the bugs I'm trying to address with this PR. I feel a little bad roping the scalebars into the buffer fix but these issues highlight each other and I wanted them fixed at the same time.

klassenjs:
Scale line looks good.  Probably should document switching it to metric (even though it appears trivial in app.js).  Buffer looks much better.  A buffer of 5280ft from a point on a property line corner lands on property lines rather than overshooting it.  I still stress that this is a (much better) approximation, but 3857 is still not a good projection to use for accurate measurements.

Old:
old


New:
new

klassenjs:
#362 is also affected by this.

PR #372, opened 2019-04-16
WIP: Fix custom projection coordinate display

klassenjs:


theduckylittle:
@klassenjs - This is :+1: from me once my fix is merged in. Not sure if that will mean opening a freshly rebased PR or not.

PR #382, opened 2019-04-30
Update install_on_windows.md

brentfraser:
Add note on CORS


klassenjs:
Question: Do we want to handle this with CORS or with the proxy in `webpack.config.js` (or both options)?

theduckylittle:
I think both is the best option. If anything having MORE docs for people who may fall into similar issues with CORS is a good idea.

Issue #373, opened 2019-04-16
Search fails immediately after selection


chughes-lincoln:
On my geomoose 3.3 setup and the demo from geomoose.org, the search fails if you try it immediately after starting a selection without finishing the selection (ie works if you click select, input a geometry and hit Go, and then try a search; fails if you click select and then try to do a search). 

Also seems to work if you use another tool in between (ie hit select, then identify or measure something, then search something) or search again immediately after it fails.

It gives me the error 'Uncaught TypeError: Cannot read property 'parse' of undefined' when it fails.

chughes-lincoln:
This now works, but the selection itself is broken. Error: "Uncaught (in promise) TypeError: Cannot read property 'split' of undefined"

theduckylittle:
#394 provided a better fix and is released.

PR #389, opened 2019-05-09
Fix state bug with switching services

theduckylittle:
ServiceForm expected to be re-created when switching
services, this creates a better test and resets the values
to ensure they don't carry from incomplete service to incomplete service.


theduckylittle:
refs: #373

Issue #329, opened 2018-10-02
Add padding around catalog/tabs


klassenjs:
The "folders" are right up against the edge and sometimes look odd.

Looks like this needs to be applied to the div inside of the tab-container div otherwise the tab content gets shifted off the other edge instead of line-wrapped at the appropriate point.  1-2px seems enough.

chughes-lincoln:
I added this code to my site.css to overwrite the geomoose.css, which worked:

```
.catalog.searchable {
  padding-top: 40px;
  padding-left: 4px;
}

.service-manager {
  padding-left: 4px;
}

.catalog.favorites.flat {
  padding-left: 4px;
}

.catalog.visble-layers.flat {
  padding-left: 4px;
}
```

Also, the div for the visible layers is misspelled - "visble"

PR #395, opened 2019-05-15
Fixed spelling, added padding.

theduckylittle:
refs: #329

### DIFF
--- a/src/gm3/components/visibleLayers.js
+++ b/src/gm3/components/visibleLayers.js
@@ -60,7 +60,7 @@ class VisibleLayers extends React.Component {
 
         return (
             <Provider store={this.props.store}>
-                <div className="catalog visble-layers flat">
+                <div className="catalog visible-layers flat">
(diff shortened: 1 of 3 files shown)


src/gm3/components/visibleLayers.js
```diff
@@ -60,7 +60,7 @@ class VisibleLayers extends React.Component {
 
         return (
             <Provider store={this.props.store}>
-                <div className="catalog visble-layers flat">
+                <div className="catalog visible-layers flat">
```
klassenjs: Are we sure people aren't using the mis-spelled version in their custom CSS?  It worries me that this bug-fix might be a breaking change.  Should we include both class names for now?  Would that even work?

klassenjs: It looks good other than that.

theduckylittle: We can leave both in there. Pushed a new commit.

Issue #408, opened 2019-11-01
Select by Box


chughes-lincoln:
It would be really useful to add the ability to select features using a box (or to get feedback about how this could potentially be added). 

The main issue for me with GeoMoose 3 is that it isn't quite as user-friendly as GeoMoose 2. If our users have to click 4 times to create a polygon rather than just clicking and unclicking to create a box, they may just not use it rather than sucking it up and putting in the extra effort.

theduckylittle:
@chughes-lincoln  - After this is merged, I would greatly appreciate a follow up PR with some docs.

@klassenjs - Are we missing a 3.5? This may be a 3.5 thing... though I've sometimes considered "forward ports" to be "bug fixes" as from the user perspective they are "Broken" from 2.x

klassenjs:
I'm fine calling it 3.5. 

Personally, I'd expect minor bumps to be the most common.  It just means we added something and did our best not to break anything.  

We technically could release a 3.4.1 with the fixes only, but all that would be is 3.5 w/o Box support, which isn't something expect would be worth the effort.

tchaddad:
Works well - I think this will be a popular one...

PR #409, opened 2019-12-05
Add Box drawing

theduckylittle:
Adds the box drawing tool to the set of options
for Select.

refs: https://github.com/geomoose/gm3/issues/408


klassenjs:
Appears to work as advertised.

A couple of notes for the future: 
* Do we need to think about the ordering of the draw tools?  To me, I tend to think of the natural ordering as Point, Line, Box, Polygon, but I don't have any solid reason for that order.  I'm also wondering if they should be more iconic than just text.

* It looks like there is an unused include in `serviceForm.js`
```
gm3/components/serviceForm.js
  35:10  warning  'jsonEquals' is defined but never used  no-unused-vars
```

Issue #407, opened 2019-10-30
Select - autoGo breaks buffer


chughes-lincoln:
In order to buffer, you need to do a search or a selection, then do another selection, clicking on whichever polygon you want to buffer. However, with autoGo = true, it immediately does the query again when you click on the polygon, without letting you set the buffer distance.

You also can't get around this by adding the buffer distance first because anytime you change the buffer distance it does the query again (i.e. if you're trying to buffer by 50ft, it runs the query when you enter a 5).

klassenjs:
What was the expected behavior?

Submitting on entering the "5" seems like unexpected behavior to me.  However, I am puzzled on how this is supposed to work.  I thought the intent of autogo was to allow the site implementor to short circuit the user needing to click "Go" to start the service in the case of simple services.  In a service that takes multiple pieces of user input, some potentially optional if the user choses to use the default values, how is it supposed to reliably determine when the user is done entering their selection and ready to submit their query?

There are other thing than edit text/select boxes before submitting the query too.  For instance, a user may want to modify the feature selected on the map, or maybe they mis-selected an existing feature and want to try again.  Maybe they want to select a feature from another layer (results or not), and then edit that feature and change the buffer until they get what they wanted.

chughes-lincoln:
AutoGo and Buffer/Modify seem to be mutually exclusive to an extent. For those options I don't know how you could configure autoGo in a way that would be useful like you mentioned. In a perfect world it would be nice to set up autoGo for certain options. i.e. if you have draw point, line or polygon selected it 'autoGoes', whereas if Multi-Point, Feature from Results, or Modify selected you have to click GO to run the query.

I know autoGo would definitely be useful (while still being able to buffer or modify polygons), as in our GeoMoose 2 setup, you can do a selection in 2 clicks which is very user-friendly and we don't want to lose that.

klassenjs:
You could configure a version of the select service without buffer/modify/etc. available and set that to autogo.  I'm not sure if all this can be overridden at the app.js level or if a copy of service.js would be needed.

As is, even if you just draw a point the user still has the option to move it/buffer it afterwards which seems incompatible with the idea of autogo.

theduckylittle:
AutoGo is _really_ meant for use with Identify and its ilk. Auto-Go makes the assumption that there are no other user interactions than the drawing. It is possible to add additional geometry options to Identify. Once the Box drawing PR is merged then it would even be possible to do a Box identify.

[...]

PR #410, opened 2019-12-05
Prevent autoGo triggering invalid state

theduckylittle:
Require inputs to be filled out (including buffer)
before submitting a form.

A console.error will appear when this is misconfigured.

refs: https://github.com/geomoose/gm3/issues/407


klassenjs:
Tested by setting `this.autoGo = true;` for the select service and verifying the error appears.  Then, testing Identify works as it used to (as a real autoGo service).

Issue #406, opened 2019-09-21
Trying to print a map containing WMS layer triggers CORS error


tchaddad:
GM 3.4.0 - printing a map with a rest service works fine, but the same service in WMS form is resulting in a CORS error message.

I tested with WMS from Mapserver, and ArcServer. 

This example is from a USGS service and drops right into the Demo:

```
<!-- ---------------- map sources --------------------- -->
<map-source name="usgs_wbd_wms" type="wms" >
	<url>https://hydro.nationalmap.gov/arcgis/services/wbd/MapServer/WMSServer?</url>
	<layer name="1"/>
	<param name="cross-origin" value="anonymous"/>
</map-source>

<map-source name="usgs_wbd_esri" type="ags" >
	<url>https://hydro.nationalmap.gov/arcgis/rest/services/wbd/MapServer</url>
	<layer name="0"/>
	<param name="cross-origin" value="anonymous"/>
</map-source>

<!-- ---------------- catalog --------------------- -->

<group title="Printing Test" expand="true">
	<layer title="WBD - ESRI REST" src="usgs_wbd_esri/0" metadata="true" legend="false" fade="true" unfade="true"/>
	<layer title="WBD - WMS" src="usgs_wbd_wms/1" metadata="true" legend="false" fade="true" unfade="true"/>
</group>
```

klassenjs:
I just tried this myself.  I see the error message, but I don't think it is CORS.  The headers look right to me and I am not seeing a CORS Blocked message on the console.  I also see the image request completing successfully in DevTools.  I suspect that it is having trouble parsing the PNG.

I added a `console.error(error);` line to printImage.js:63 I see a:
```DOMException:The Operation is Insecure``` error line 87.

Stack:
```
getImage printImage.js:61:20
mapRendered printImage.js:72:46
mapRenderedCallback printImage.js:94:49
boundListener events.js:41:11
dispatchEvent Target.js:99:12
renderFrame_ PluggableMap.js:1256:4
PluggableMap/this.animationDelay_< PluggableMap.js:189:6
```

tchaddad:
yeah CORS didn't really make sense, but I didn't have a ton of time to dig deeper, so thanks for confirming with a  bit more info.

tchaddad:
I tested this bug in all the versions back to 3.0, and it looks like it was in all of them...

tchaddad:
Also tested with requests for different image formats from the WMS, and it didn't seem to make a difference.

[...]

PR #411, opened 2019-12-05
Fix for cross-origin WMS requests

theduckylittle:
`cross-origin` properties were not being honoured
on WMS layers.

refs: https://github.com/geomoose/gm3/issues/406


klassenjs:
I wish we could have tests for stuff like this (Does map, query, and print work with all layer types same-origin, cross-origin, with different server software, etc?).   But, that just seems like an impossible amount of stuff to carry with just for the tests (and I really don't want to depend on external internet services to run the tests).

Issue #398, opened 2019-05-17
Add scaleline to printing


theduckylittle:
As we dropped the mapserver raster scalebars, we should add a version of the OpenLayers-style scale line to the print outs.

tchaddad:
Any chance of this one making it into 3.5?

theduckylittle:
@tchaddad - Looking to try and get this in. I have a PR that @klassenjs and I have differing perspectives on 'working.' We may need more testing and feedback.

tchaddad:
Sure thing.

Saw both your tests and noted that @theduckylittle tested miles, while @klassenjs tested kilometers. 

Could there be a units conversion problem somewhere that is affecting the results of @klassenjs 's test?

tchaddad:
Works for me now, for all print options

tchaddad:
Closing.

PR #415, opened 2019-12-10
Add a scaleline to the print out

theduckylittle:
Draws a scaleline on the PDF when configured
to do so. The settings are the same ones used
to render the scaleline for the map.

refs: #398

### DIFF
--- a/examples/desktop/app.js
+++ b/examples/desktop/app.js
@@ -32,12 +32,14 @@
 var app = new gm3.Application({
     mapserver_url: CONFIG.mapserver_url,
     mapfile_root: CONFIG.mapfile_root,
+    /*
     map: {
         scaleLine: {
             enabled: true,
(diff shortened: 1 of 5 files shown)


klassenjs:
It looks like the scale is off.  It says 2km, but is really closer to 3km.

image

(Printed, saved, opened in GIMP, relocated the scalebar over the grid).

theduckylittle:
Uh... I'm not seeing the same issue (looking at a 2mi grid).

duck-2mi-grid

theduckylittle:
Another example that 'looks right'...

image

klassenjs:
Figured out how I broke it: Change the print resolution.

klassenjs:
Also found a new bug that confused me in the first place. #421

klassenjs:
Looks good.

PR #416, opened 2019-12-11
Remove catalog sources for old MapServer scalebars

tchaddad:
I'm pretty sure these are all the old scalebars, so should be deleted from the demo mobile Mapbook...

### DIFF
--- a/examples/mobile/mapbook.xml
+++ b/examples/mobile/mapbook.xml
@@ -41,26 +41,6 @@
         </layer>
     </map-source>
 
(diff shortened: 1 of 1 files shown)


klassenjs:
@theduckylittle Is there any reason the mobile and desktop examples are using separate mapbooks?

theduckylittle:
@klassenjs  - To keep things independent and show how someone might have 'multiple' applications.

klassenjs:
@theduckylittle - Ok.  I also think it is interesting to show people they can have multiple interfaces into the same application.

PR #422, opened 2019-12-27
Updated Package.json for publishing

theduckylittle:
1. Changes the package name to @geomoose/gm3
2. Removes downloading dev-only packages. Since we distribute a `dist`
   directory, there is no need to also download all the dev deps.


klassenjs:
Note: This change requires updates on the build server to accommodate the new tarball name produced by `npm pack` due to the changed package name.

Issue #421, opened 2019-12-18
Print dialog gets out of sync with itself


klassenjs:
1. Print -> Set Resolution to Higher or Highest -> Print
1. Print -> Resolution is set to Normal, but actual map resolution is still as set the last time.

To get it to actually go back to normal set it to something other than Normal, then set it back to Normal.

Same thing happens with the Page Layout option.

PR #424, opened 2019-12-27
Fix state management issues with print dialog

theduckylittle:
Using the "refs" is both evil and broken. This changes that.

refs: #421


klassenjs:
Title and resolution are behaving.   Layout is still acting up.

theduckylittle:
@klassenjs - Found the one line fix. This should be good to merge.

Issue #348, opened 2019-01-30
Unicode characters when typing title while using print service 


stathisar:
When typing the title name of the map while using the print service, if the characters are Greek or Cyrillic they are not displayed in the pdf. It seems that there is a lack of support of Unicode characters

klassenjs:
Confirmed.
Looks like codepoints less than work (such as `é` U+00E9).

`ͻ` (U+037B) gets rendered as `{`
`؆` (U+0606) doesn't get rendered at all.

klassenjs:
Looks like support was added in jsPDF 1.4+ and also requires an appropriate font (I'm not sure if whatever font we are using counts for that or not).

More info in: https://github.com/MrRio/jsPDF/issues/12

theduckylittle:
In a local branch, I've upgraded jsPDF to 1.5.x but that didn't fix the problem as there does need to be more work around font support.  I'll need to keep this issue open as I suspect it may lead to needing to do some documentation on how to properly add a new font to the PDF code.

PR #426, opened 2019-12-27
Fix for print issues with international characters

theduckylittle:
Adds the Noto Sans font as the default which
is now including a full set of cyrillic, latin,
and greek characters.

refs: #348


theduckylittle:
I'm not thrilled with this current solution as it adds 1.8MB to the build output.

Possible follow up thoughts:
1. Drop some of the font styles (say "bold italic")
2. Async load the font files (has issues with deploy and predictable locations)
3. Use the web-font features of jsPDF but that just requires loading them via the style sheet before the page loads. Has the same effect on download but makes `geomose.min.js` smaller.

klassenjs:
Instead of embedding a font inside the geomoose.js bundle, can we let the user configure a URL to a font to use (in app.js or config.js)?

My thinking is this would delay downloading the relatively large font until someone actually prints (this could be particularly important on mobile where size matters and printing is unlikely).  It also lets the admin choose a font that fits their needs, be that character set coverage or just style.

We can include Noto (or whatever) in the demo as an example of how this works, ideally it would fall back to the current behavior if the font isn't set (as to not surprise existing installs).

elil:
1.8MB is not that big in 2019

klassenjs:
An additional 1.8MB of JavaScript is a lot to add to every page load (both download and parsing) just in case someone wants to include a title in a PDF printed map that contains some non-Latin-1 characters.  But worse, this 1.8MB doesn't even solve the problem as it only covers _some_ of the potential i18n characters someone might want to use.

We would have to also include additional fonts to cover CJK or Arabic for instance, and there are many others too.  We would then have to figure out which of those fonts to use for a given combination of characters.

Thus, I think it is best to allow the site admin choose a font that works best for the needs of their site (without rebuilding GeoMoose to do it).

klassenjs:
Work in progress/proof-of-concept for letting the admin configure the font.  This has some really ugly spots in it yet, but it working for me.

As it is so ugly, I don't want to commit it on top of this PR, I don't want to make competing PRs, and I'm not sure how to make a pull request to a pull request so:

klassenjs/gm3@2072184b2a2548f58544e6bc1b316af9f98d9d01

klassenjs:
This appears to work.  There is one oddness where the font.js file is being generated twice and in different ways by webpack.  This makes the npm package larger than it need be.  This potentially wastes some space on the server, but shouldn't effect the end user.

I'd still like to look at making this cleaner/smaller later, but I think it is good enough for now.

PR #427, opened 2019-12-30
Updated Build System

theduckylittle:
1. Removes all the grunt tasks and switches to pure-NPM with
   some new tasks.
2. Drops the dual webpack config files in favor of one that
   can honor environment variables.
3. Drops "debug.html" from the desktop example as geomoose.min.js
   can be served from the webpack-dev-server. It's just a name
   the dev-server version still includes debugging symbols.


klassenjs:
Results as of 95de08b:
- serve, build, pack, test, cover, test w/test_deps (wfs+map) successful on Linux/Node 10
- serve, build, pack successful, test fails with jest errors on Win/Node 10

klassenjs:
- verified this builds more or less the same thing as master (keeping in mind there were many dependency versions that updated).

Issue #419, opened 2019-12-17
Results grid covers scalebar (and map)


chughes-lincoln:
When you do a selection on the demo, the results grid pops up covering up the lower third of the map, including the scalebar. It still covers the scalebar even if you minimize the results table.

chughes-lincoln:
This looks great. I also really like how the map also adjusts when you toggle the results grid, rather than just fixing the scalebar.

PR #429, opened 2020-01-04
Adding skeleton to normalize CSS changes + use flex layout

theduckylittle:


klassenjs:
Overall, looks good.

This is a little ugly (some controls are crowded by boarders) with old examples and new css but it technically works.  Practically, I think people will need to update their index.html.  I'm not sure if this is enough to warrant a major version bump or not.

klassenjs:
Cool, now the old apps still look good with the updated library.

We need to remember to tell people they might want to update their apps with the new CSS, etc.

klassenjs:
I think this takes care of #419

PR #431, opened 2020-01-05
Fix issue with disappearing tool interaction

theduckylittle:
When "End Drawing" is clicked it sets the interactionType to null.
However, when re-clicking the same service, this was causing a weird
null-state to be displayed to the user. They had started the service
but no drawing tool would be selected.

This restores the drawing tool to the default no matter what.

### DIFF
--- a/src/gm3/components/serviceManager.js
+++ b/src/gm3/components/serviceManager.js
@@ -356,6 +356,7 @@ class ServiceManager extends React.Component {
 
             if (service_def &&
                 service_def.tools &&
+                this.props.map.interactionType == null &&
                 service_def.tools.default !== this.props.map.interactionType) {
                 // TODO: The last tool state for a service should be saved so
                 //       as much interaction state as possible can be restored.
--- a/src/gm3/components/serviceManager.js
+++ b/src/gm3/components/serviceManager.js
@@ -353,6 +353,16 @@ class ServiceManager extends React.Component {
         } else {
             const service_name = this.state.lastService;
             const service_def = nextProps.services[service_name];
+
+            if (service_def &&
+                service_def.tools &&
+                service_def.tools.default !== this.props.map.interactionType) {
+                // TODO: The last tool state for a service should be saved so
+                //       as much interaction state as possible can be restored.
+                // restore the default tool when re-clicking the service.
+                this.props.changeDrawTool(service_def.tools.default);
+            }
+
             // if this service has 'autoGo' and the feature is different
             //  than the last one, then execute the query.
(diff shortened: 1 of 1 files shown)


klassenjs:
It looks like there are some issues here.

Measure tool fails (with type error i is undefined) if change from line to either point or polygon.  Measure tool fails the same when double click to end the line.

Select tool goes back to polygon after every complete geometry.  Multi-point doesn't work at all because it considers each point a completion, so you can't add the next point (it switches back to polygon and starts a polygon instead).

klassenjs:
```js
TypeError: _service_def is undefined   2 geomoose.min.js line 11057 > eval:504:21
    UNSAFE_componentWillUpdate serviceManager.js:357
    React 10
    performSyncWorkOnRoot self-hosted:1053
    flushSyncCallbackQueueImpl React
    unstable_runWithPriority scheduler.development.js:697
    React 6
    onStateChange Redux
    onStateChange self-hosted:1007
    dispatch Redux
    addSelectionFeature map.js:856
    feature_change_fn map.js:897
    boundListener events.js:41
    dispatchEvent Target.js:99
    addFeatureInternal Vector.js:319
    addFeature Vector.js:290
    finishDrawing Draw.js:879
    handleUpEvent Draw.js:578
    handleEvent Pointer.js:132
    handleEvent Draw.js:524
    handleMapBrowserEvent PluggableMap.js:931
    boundListener events.js:41
    dispatchEvent Target.js:99
    handlePointerUp_ MapBrowserEventHandler.js:166
    boundListener events.js:41
    dispatchEvent Target.js:99
    fireNativeEvent PointerEventHandler.js:393
    pointerUp NativeSource.js:64
    pointerUp self-hosted:1009
    eventHandler_ PointerEventHandler.js:199
    boundListener events.js:41
```

klassenjs:
I'm really not sure what this fixes.

No change:

1) Select Features->Draw Polygon->End Drawing => no tool selected
2) Select Features->Draw Line (or any of them)->End Drawing->Identify->Select Features => polygon selected

Before:

3) Select Features->Draw Multi Point->Click on map => multi point tool selected
4) Select Features->Draw Polygon->draw on map->Go->Select Features => no tool selected (unexpected)

After:

3) Select Features->Draw Multi Point->Click on map => Resets to polygon (unexpected)
4) Select Features->Draw Polygon->draw on map->Go->Select Features => polygon selected

So, this fix seems to have nothing to do with clicking "End Drawing", but more to do with what happens when clicking Go.  My thinking is there should be an additional test that only resets to the default tool when `map.interactionType` is null.

Issue #434, opened 2020-01-08
Print Dialog Layout


klassenjs:
The contents of the print dialog no longer fit well in the dialog pane since #429.  This is an issue because the user may miss that the preview map is available (and important given that we use the generation of the preview map is the notification to the user that the application is ready for them to click Print).

PR #435, opened 2020-01-09
Fix print preview image being buried

theduckylittle:
Adds additional CSS support to the Modal's render body method.


klassenjs:
Fixes #434

Before:
image


After:
image

PR #447, opened 2020-01-16
CSS improvement to the tab-toggler

tchaddad:
Restoring the background to the tab-toggler. Think this will fix the CSS report I had. Tested on FF and Chrome.

### DIFF
--- a/examples/desktop/tabs.css
+++ b/examples/desktop/tabs.css
@@ -45,10 +45,15 @@
     z-index: 1000;
     transform: rotate(0deg);
     transition: left ease 500ms, transform ease 500ms;
+    background-color: white;
+    border: solid 1px black;
+    border-radius: 0px 10px 10px 0px;
+    padding: 1px;
+    margin-left: 2px;
 }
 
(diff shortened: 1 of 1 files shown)


theduckylittle:
This actually changes a lot of the CSS. What is the actual bug?

tchaddad:
The bug was about the tab toggler disappearing, and then not being able to get it back. I couldn't reproduce.

While I was in there, I noticed that the tab behind the tab toggler had been removed, so I was suggesting putting it back with this PR. 

Looks better to me, but you can also just leave that for implementers if you want to keep things lean.

theduckylittle:
I want to keep this PR open for now. There is a chance that I'll get around to having a real adjuster in the next release making the tab toggler a legacy component.  But I'd like to think on it for 3.5.1 or 3.6.

@klassenjs  - This is non-blocking to the 3.5.0 release.

theduckylittle:
I have a smaller tweak that leaves more of the styling updates in place, so closing this one.

Issue #449, opened 2020-01-17
demo could benefit from multiple="false" (radio buttons) in catalog


elil:
Right now the demo lets users add multiple basemap layers at a time, while this could be useful in the case of transparency, the more common case will be inefficiencies with multiple basemaps on and confusion over why the desired layer is not displaying.  

Looking at #351, this has come up before.  Do we want to look at it again?  Maybe a restricted option for 1:1 cases.  Maybe we could come up with a mapbook validator that would check things like this?

bblammo:
hmm, this is an interesting one.

We've run into this from time to time, butnit hasn't really been a big
deficiency because we have a smaller and faurly static user group.  I can
see thisnbeing more of a problem with occasional use.

What about using some sort of status bar with circles and numbers inside
indicating hiw many of which type of layer are on/active.

alt tags for describing what each number represents.

bobb

chughes-lincoln:
I think when start having larger user groups that aren't as accustomed to GIS it will become more of an issue. Right now our site has several high resolution imagery layers plus a couple lidar hillshades which will use a lot of resources if users don't know to turn off one layer to switch to a layer below it (plus the inconvenience of not being able to click once to choose which 'basemap' layer you want)

theduckylittle:
refs: PR #451

elil:
Radio buttons appear to work in the demo, closing.

PR #451, opened 2020-01-21
Restoring radio-button groups

theduckylittle:
- Adds the style to the catalog for radio buttons
- Honors the old multiple=false from GM2.x

### DIFF
--- a/examples/desktop/mapbook.xml
+++ b/examples/desktop/mapbook.xml
@@ -503,7 +503,7 @@
             <layer title="1 mile grid" src="grids/grid_1mile"/>
         </group>
 
-        <group title="Backgrounds" expand="true">
+        <group title="Backgrounds" expand="true" multiple="false">
             <layer title="Reprojected Regional Aerial" src="lmic/mncomp" show-legend="false" legend="false" fade="true" unfade="true"/>
 
             <layer title="OpenStreetMap" src="openstreetmap/osm_mapnik" legend="false" fade="true" unfade="true"/>
--- a/examples/desktop/mapbook.xml
+++ b/examples/desktop/mapbook.xml
@@ -504,6 +504,7 @@
         </group>
 
         <group title="Backgrounds" expand="true" multiple="false">
+            <layer title="No background" src="blank/blank" status="on" />
             <layer title="Reprojected Regional Aerial" src="lmic/mncomp" show-legend="false" legend="false" fade="true" unfade="true"/>
 
             <layer title="OpenStreetMap" src="openstreetmap/osm_mapnik" legend="false" fade="true" unfade="true" />
--- a/examples/desktop/mapbook.xml
+++ b/examples/desktop/mapbook.xml
@@ -425,7 +425,7 @@
     </map-source-->
 
     <map-source name="openstreetmap" type="xyz">
-        <layer name="osm_mapnik">
+        <layer name="osm_mapnik" status="on">
             <attribution><![CDATA[
                 &copy; <a href="http://www.openstreetmap.org/copyright" target="_blank">OpenStreetMap</a> Contributors.
             ]]></attribution>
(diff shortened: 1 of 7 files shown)


klassenjs:
My memory may be failing me here, but I thought we very purposely didn't include this in GM3 after a lot of consideration.  I think it had something to do with bad interactions with attaching more than one source to a catalog layer.  I'd go dig, but I'm very tired right now.

elil:
See #449 that has some history.  Also adds references for commit.

chughes-lincoln:
I definitely think it would be useful to allow this to be configured (with a doc explaining the bad interactions if that's the case). Being able to prevent users from turning on multiple basemaps is very helpful, especially when dealing with users that aren't GIS professionals.

theduckylittle:
The _big_ problem is that we allow a single `<layer>` in the `<catalog>` to represent multiple `<map-source>`s. This could, potentially, allow an administrator to create a state where only one of the layers in a checkbox is on and others are not. With a checkbox we can sorta ignore this. We could even add a "partial" CSS if we really felt like communicating that to the user. Radio buttons are more problematic. If the administrator configures things wrong the user could be presented with two radio buttons that are on. That is super disorienting. 

That said, I'm increasingly okay with letting admins hurt themselves a little and need to debug.

klassenjs:
Historically there were other ways they could hurt themselves with this.

Adding/removing the layer from the hashtrack

Having one multiple block force the layer off while another forces it on

Possibly some interaction with visible layers tab

...

I guess I'm ok with it so long as there is a rule we can include in the docs and warn people (console.warn/error) if it is violated.  I just don't want to deal with the frustration on both sides when people configure something that is logically inconsistent and then blame us when it fails.


So, what is the rule???

Any source/name used within a multiple=false group must only be used once (across the whole mapbook)?

I mean someone might find some use case where violating the rule works for them and/or the resulting breakage doesn't bother them, and if they knowingly make it that decision then I'm fine with that, but they are on their own at that point.

chughes-lincoln:
I noticed a couple issues when I tried to incorporate this into my GeoMoose build. (I haven't noticed the above issues but that's because I only have one `<map-source>` for each `<layer>` in the catalog)

1). I haven't been able to get this working for nested groups.

	<group title="Aerial Photos" expand = "true" multiple = "false">
		<group title="Highest Resolution Imagery" expand = "true">
      <layer title="2018 Aerial Imagery" src="photo_mapserver/Lincoln2018_map" status="off" legend="false" show-legend="false" fade="false" unfade="false" tip='1 foot resolution'/>
			<layer title="2013 Aerial Imagery" src="photo_mapserver/Lincoln2013_map" status="off" legend="false" show-legend="false" fade="false" unfade="false" tip='1 foot resolution'/>
			<layer title="2007 Aerial Imagery" src="photo_mapserver/Lincoln2007_map" status="off" legend="false" show-legend="false" fade="false" unfade="false" tip='1 foot resolution'/>
		</group>
		<group title="Lower Resolution Imagery" expand = "true">
			<layer title="2016 Aerial Imagery" src="OregonAerials/Framework_Imagery_Mosaic2016" legend="false" show-legend="false" fade="false" unfade="false" tip='1/2 meter resolution'/>
			<layer title="2011 Aerial Imagery" src="photo_mapserver/Lincoln2011_map" status="off" legend="false" show-legend="false" fade="false" unfade="false" tip='1 meter resolution'/>
		</group>
		<layer title="None" src="blank/blank" legend="false" fade="false" unfade="false"/>
	</group>

i.e. in the above code, setting multiple = false for the Aerial Photos group doesn't change anything, but doing the same for either the Higher or Lower resolution groups works fine.


2). When you've set multiple to false, there's no way to turn a layer off. I had to add back in a 'blank' `<layer>` in the catalog as above and add a blank `<map-source>` like we did in geomoose 2 (as well as a blank gif)

      <map-source name = "blank" type = "wms" tiled = "false">
        <url>./images/blank.gif </url>
        <layer name = "blank" status = "on"/>
      </map-source>

theduckylittle:
Not sure I'm comfortable making the "multiple" an inherited property. That would require traveling up the catalog tree quite a bit instead of simply up to the parent. Generally, radio-list operators are flat and don't go down the branches.

Adding a 'blank' layer type is a good idea for this scenario again. Adding to my TODO list.

theduckylittle:
This is also non-blocking for 3.5.0.

PR #453, opened 2020-01-21
Update selection options doc

chughes-lincoln:
Gives additional guidance about how to further configure select options without causing additional errors


chughes-lincoln:
Guidance for configuring autoGo without errors

theduckylittle:
This looks great! Thank you so much!

PR #457, opened 2020-01-23
Update add-search.md

chughes-lincoln:
Added guide to configuring a search that gives a warning if you don't input anything

This assumes they have a dataset to modify the demo to


theduckylittle:
Can we break this out into a separate file? With the way the how-tos are structured, I think the preference is to keep them short and focused to one topic.

chughes-lincoln:
I broke it apart into a couple files and updated the index to also point to the new doc. Let me know if there are any other changes I could make.

theduckylittle:
Looks great!

Issue #95, opened 2017-05-29
Grid results and tab results behave differently


elil:
Hovering over grid results highlights that particular parcel in the selection.  Hovering or click on results in the super tab does not do this.  It would be better if the grid and tab behaved in a similar manner.  

Hover could perhaps also be clicking instead.

klassenjs:
I think this is up to how the admin configures the layer in the mapbook.  (Not that the demo couldn't be configured better.)

In the demo the results tab feels kind of redundant to the grid as currently configured.

elil:
Yes, let's configure the demo well to get people off to a good start.  

I'd like the results and grid to match and be synchronized.  Click/Hover/Select in either the results or grid and both it and the other gets highlighted in red.  The feature on the map is already nicely doing this (only from grid).  

Another approach could be that when grid is maximized, the results tab only shows the highlighted result.  

#124 might help this some too.

theduckylittle:
@elil - I've done some improvements. Good enough for 3.0.0? Got any more suggestions?

elil:
I like the summary information!

Is there a scenario in which the layers will be a number other than 1 (for
select)? (I see Identify which does multiple layers uses the same thing and
makes sense).

This could work for 3.0.0.  I'd like to see the tab have identify-like
template information for the individually selected/highlighted grid row.
That way the white space does something productive.  It might just be me
favoring the familiar too...

chughes-lincoln:
I was able to add the ability to highlight tab results while hovering over them. All you need to do is modify the relevant section in the search template with the highlight section of the parcel-row.html
See where I've made the modifications followed by the entire modified template: https://github.com/geomoose/gm3/blob/master/examples/desktop/mapbook.xml#L74

`<template name="search"><![CDATA[
                <div class="search-result"
                  onmouseenter="app.highlightFeatures({'PIN' : '{{ properties.PIN }}'}, true)"
                  onmouseleave="app.clearHighlight()"
                  >
                    <div class="search-label">
                        {{ properties.OWNER_NAME }}
                    </div>
                    <div class="search-action">
                        <div style="padding: 2px">
                            <a onClick="app.zoomToExtent([{{ properties.boundedBy | join }}])" class="zoomto-link">
                                <i class="fa fa-search"></i>
                                {{ properties.PIN }}
                            </a>
                        </div>
                    </div>
                    <div class="search-address">
                        {{ properties.OWN_ADD_L1 }}<br/>
[...]

PR #458, opened 2020-01-26
Add highlighting to search results

theduckylittle:
refs: #95


theduckylittle:
Non blocking for 3.5.0, could be included if tested quick, or saved for 3.5.1.

Issue #456, opened 2020-01-23
Search - Field Required


chughes-lincoln:
If you select the search tool and click GO without entering anything, it warns you that you need to complete a field (when you've configured the search to throw the warning). (See https://github.com/geomoose/gm3/issues/316). However, if you hit enter instead of GO you avoid this warning. 

I would guess that users in general are more accustomed to searching by hitting Enter rather than clicking a search button (i.e. searching on the web, google maps, etc...) and would run into this and unintentionally search the entire dataset.

chughes-lincoln:
Works great, I think this will help prevent users from using the service incorrectly. Feel free to close it, I'm leaving the issue open right now just because the fix hasn't been pulled yet.

theduckylittle:
Merged. So closing.

PR #459, opened 2020-01-27
Fix for keyboard-bypass of field checks

theduckylittle:
Centalizes submit logic in service form for
the button and the keyboard.

refs: #456


theduckylittle:
Tested by @chughes-lincoln, merging.

PR #460, opened 2020-01-27
Update the measure tool how to to include initialUnits

theduckylittle:
As queried by chughes.


chughes-lincoln:
Works as expected, thanks!

PR #461, opened 2020-01-28
Add 'hidden' Input Type

theduckylittle:
This is a future port to allow hidden values in
the service definition.


klassenjs:
How does that help when a service is custom local javascript and not a XHR call?

PR #465, opened 2020-01-29
Add Plugin Support

theduckylittle:
This might _never_ get merged or expect the work to live on elsewhere.

This is an example of how one might build a react-based plugin for GeoMoose.

### DIFF
--- a/docs/howto/add-a-plugin.md
+++ b/docs/howto/add-a-plugin.md
@@ -1,0 +1,84 @@
+# How-to add a Plugin
+
+Plugins allow adding code into your GeoMoose application at a
+deeper level. They require familiarity with React and Redux
+but expose many of GeoMoose's inner-workings.
+
+## 1. Write the plug-in
+
+Plugins are given the React, ReactDOM, and redux store.
+They are expected to return a component. The following example
+uses a functional component:
+
+`examples/plugin/src/hello.js`:
+```
+// If using the default settings for @babel/react-preset then
+//  keeping the names React and ReactDOM are important.
+const HelloWorldComponent = ({React, ReactDOM, store}) => (
+    <div>
+        This application has { Object.keys(store.getState().mapSources).length } map sources.
+    </div>
+);
+```
+
+This example is written using ES6. It features the JSX syntax. It is not
+the scope of this How-to describe those concepts, however, the GeoMoose repository
+contains all the things we need to convert this to browser-friendly ES6.
+
+```
+cd examples/plugin
+npm install
+npm run build
+```
+
+This will create `examples/plugin/hello.js`.
+
+## 2. Adding it to the application
+
+`hello.js` is a little informative panel to tell the user how many map sources have been configured.
+The following changes will add a tab to the application then add the plug-in to that tab.
+
+### Adding a tab
+
+1. Edit the `index.html` file of your application and find the *Super Tab*.
+2. The following will add a new tab to the list of tabs:
+
+   ```
+   <span id="hello-tab-tab" class="tab" onclick="showTab('hello-tab', event)">Hello Tab</span>
+   ```
+
+3. And add the new tab content:
+
+    ```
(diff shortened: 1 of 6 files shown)


elil:
This is a great way to show by example some of the possibilities!

klassenjs:
The `window.React = React` stuff scares me.

klassenjs:
But as long as we are discussing dirty hacks, can we put `React` in `gm3.Application`?

At least the `app` global is the user's discretion and doesn't magically pollute window.

And the circular references don't matter if the object is never deleted, right?

theduckylittle:
@klassenjs That has also been a concern of mine. I just refactored the code a little to _pass_ react in as a prop to the component. It requires the user do a little extra work but really would allow anyone with the motivation to really push a GeoMoose integration.

theduckylittle:
In terms of references, React will never get unloaded because of how we use it in the app.

theduckylittle:
@klassenjs -  I realized I didn't actually push my changes. Now you can see the updated version. I think this satisfies everyone's scope concerns.

klassenjs:
How does hello.js get to examples/desktop/hello.js?

theduckylittle:
I would like to change this to get rid of the change to examples directory. I can write up the how-to if you're okay with the implementation.

theduckylittle:
Updated the docs, cleaned up the changes to the desktop example

docs/howto/add-a-plugin.md
```diff
@@ -0,0 +1,85 @@
...
+2. The following will add a new tab to the list of tabs:
+
+   ```
+   <span id="hello-tab-tab" class="tab" onclick="showTab('hello-tab', event)">Hello Tab</span>
+   ```
+
+3. And add the new tab content:
+
+    ```
+    <div class="tab-content" id="service-tab"></div>
```
klassenjs, line 53: id="hello-tab"

docs/howto/add-a-plugin.md
```diff
@@ -0,0 +1,85 @@
...
+    ```
+
+### Adding the plug-in
+
+1. The plug-in will need copied from where it was just built to your application's directory. In the demo this is `examples/desktop` but your apps directory may be different:
+
+    ```
+    cp examples/plugins/hello.js examples/desktop/
```
klassenjs, line 61: plugin not plugins

docs/howto/add-a-plugin.md
```diff
@@ -0,0 +1,85 @@
...
+deeper level. They require familiarity with React and Redux
+but expose many of GeoMoose's inner-workings.
+
+## 1. Write the plug-in
+
+Plugins are given the React, ReactDOM, and redux store.
+They are expected to return a component. The following example
+uses a functional component:
+
+`examples/plugins/src/hello.js`:
```
klassenjs, line 13: plugin not plugins

docs/howto/add-a-plugin.md
```diff
@@ -0,0 +1,85 @@
...
+    </div>
+);
+```
+
+This example is written using ES6. It features the JSX syntax. It is not
+the scope of this How-to describe those concepts, however, the GeoMoose repository
+contains all the things we need to convert this to browser-friendly ES6.
+
+```
+cd examples/plugins
```
klassenjs, line 29: plugin not plugins

docs/howto/add-a-plugin.md
```diff
@@ -0,0 +1,85 @@
...
+npm install
+npm run build
+```
+
+This will create `examples/plugins/hello.js`.
```
klassenjs, line 34: plugin not plugins

docs/howto/add-a-plugin.md
klassenjs: And maybe show the before and after state.

theduckylittle:
Fixed the editing issues!

PR #466, opened 2020-02-04
Docs for select from list

chughes-lincoln:
Added docs explaining how to configure select tool that selects from list of layers (or single layer) rather than all selectable layers.


theduckylittle:
This is great! Thanks again!!!

PR #471, opened 2020-03-12
Added header and footer templates to search results

chughes-lincoln:
I added header and footer templates to the search service in the same way as the select service. To take advantage of this you would have to add 'search-header' templates to the map-source in the `mapbook.xml` just as you would for the select service. Feel free to incorporate this, modify it, or throw it out.


theduckylittle:
Thanks for the add! Nice change!

Issue #438, opened 2020-01-09
Odd select behavior when default is set to Feature from Results


chughes-lincoln:
If I set the default selection option to Feature from: Results it doesn't truly retain that as a default (it appears to, but you need to click on that option to actually select the feature you wish to buffer).

To recreate: Modify select.js so that the default tool is 'Select'. Do a search to get some results. Click the select tool (note that Feature from: Results is checked. Click on result on map to see that nothing happens. Click on Feature from: Results as you would if it weren't selected. Click on result on map to see that it now works.

It's possible this was fixed with recent commits in the past couple days that I haven't incorporated into my gm3 build. i.e: https://github.com/geomoose/gm3/issues/428

chughes-lincoln:
Note: This wasn't fixed via https://github.com/geomoose/gm3/issues/428

chughes-lincoln:
This fixes the issue for me. Thanks!

PR #476, opened 2020-04-04
Fix default state of activeSource for select

theduckylittle:
If the Select tool is the default (pick-a-feature-from-alayer)
then it was not setting the appropriate default interactive layer.

This checks to ensure that when Select is the default it will
use the first layer that is active.

refs: #438

### DIFF
--- a/src/gm3/components/drawTool.js
+++ b/src/gm3/components/drawTool.js
@@ -72,8 +72,7 @@ class DrawTool extends React.Component {
     componentDidMount() {
         // if starting up with the select tool,
         //  ensure there is a valid active layer.
-        if (this.props.geomType === 'Select') {
-            // this.changeSelectLayer(this.props.selectableLayers[0]);
+        if (this.props.interactionType === 'Select' && this.props.geomType === 'Select') {
             const firstLayer = this.props.selectableLayers[0];
             this.setState({selectLayer: firstLayer});
             this.props.onChange('Select', firstLayer);
--- a/src/gm3/components/drawTool.js
+++ b/src/gm3/components/drawTool.js
@@ -36,7 +36,6 @@ class DrawTool extends React.Component {
     constructor(props) {
         super(props);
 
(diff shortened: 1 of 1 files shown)


klassenjs:
Is the select tool supposed to default to "From Results" in the demo?  That would seem to make it difficult to do the first round selection.

theduckylittle:
Yes, we're at a bit of an impasse there because everything is really configured by the app.js and there is only so many ways to pass info to the component right now. So while it may not be the best user experience at least this PR fixes the legitimate bug.

klassenjs:
I guess it is in `select.js`, not `app.js`.  But, why is 
```js
this.tools {
    ...
    default: 'Polygon',
    ...
}
```

being ignored now?

I get why being able to configure 'Feature From:'  as the default 'tool' might be desirable.  I don't get why it is ignoring the default (Polygon) that is configured.

theduckylittle:
Can you be more specific? I'm not seeing a bug with that. When I open up the select tool normally it does open with `Polygon` as the default drawing tool

chughes-lincoln:
The default selection option in the demo is still Polygon. However, previously if you changed the default to 'From Results' it wouldn't correctly retain that default (For example, in my build I broke the select service into a select service and a buffer service - I couldn't correctly change the default to 'From Results' in the buffer service)

chughes-lincoln:
Also, this fixes that issue on my build

klassenjs:
I'm still seeing this with 7b2313c.  New build on a different machine.

Start GeoMoose.  Open the Select Tool.  "Feature From: Results" is selected.  I'd expect Polygon to be selected.

chughes-lincoln:
I tested it again on my build and I'm seeing the same thing now (Feature From: Results is selected when Polygon is set as the default). I had Feature From: Results turned off in my select service, so it was still correctly recognizing Polygon as the default.

klassenjs:
@chughes-lincoln Thanks for verifying.  I can stop trying to figure out where my build went wrong now.

PR #477, opened 2020-04-05
Add the removal of a single feature to the demo

theduckylittle:
Now a feature can be removed from the results set from
a select. There was a state issue that required adding a
revision counter to the reducer.


chughes-lincoln:
Tested it out and it works well for me (correctly removes from map, table, tab, csv results).
You can also add this to the tab results for searches and selections (if you've set up the selection to return results in the tab).

PR #478, opened 2020-04-06
Add documentation for "or" type searches

theduckylittle:
Describes how to change the search service configuration to
support "or" searches by both:
1. Allowing specific fields to be joined by "OR"
2. Hide the specific search fields from the user.
   By using a single unified search term.

### DIFF
--- a/docs/howto/configure-search-or-fields.md
+++ b/docs/howto/configure-search-or-fields.md
@@ -1,0 +1,112 @@
+# How-to Configure "or" type Search
+
+## Allow any matching field.
+
+The following change will match any field. For example, it will match both the OWNER_NAME field and the OWN_ADD_L1 field if they both contain search terms.
+
+In the `app.js` configuration after:
+```
+    app.registerService('search', SearchService, {
+        fields: [
+            {type: 'text', label: 'Owner Name', name: 'OWNER_NAME'},
+            {type: 'text', label: 'Street/Address', name: 'OWN_ADD_L1'},
+            {type: 'text', label: 'City/State/ZIP', name: 'OWN_ADD_L3'}
+        ],
+```
+
+Add the following:
+```
+        prepareFields: function (fields) {
+            // reformat the fields for the query engine,
+            //  "*stuff*" will do a case-ignored "contains" query.
+            var query = ['or'];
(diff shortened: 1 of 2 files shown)


chughes-lincoln:
The simple search works well (searching several fields using one input box).

Searching by list also works, although as configured it looks like you would have to create a service 
that can only search a single field. Is there any way to add more fields? (i.e. by adding this to the 
above 'simple' search or by specifying more than one field in the fields parameter?). Ideally we could include this functionality for people who needed it and the people who don't need it wouldn't even know it's there.

theduckylittle:
1. Fixed the formatting issue.
2. Added docs on how to search multiple fields with a list.

klassenjs:
Did you mean to change app.js?  I'm a little worried about losing the old field validation example.  Bigger picture, I'm not sure how to best include multiple concepts in a single demo without making it hard to follow.  Are we showing all that can be done or things that make sense to do together?

theduckylittle:
Nope! That was oversight. Though, I think, arguably, one of these searches probably demos better than the default "pick a field" version. We also aren't exposing any operators other the `like` in the current demo.

I think it's fair to ask what we should evolve in the demo to make it look more current/competitive/relevant. But we can file those as follow up enhancement tickets.

klassenjs:
I wonder if we need more than one search service (even if on the same layer).

chughes-lincoln:
I made some slight modifications to add back in the field validation from the demo search: https://github.com/geomoose/gm3/pull/482

elil:
@klassenjs I think that the demo definitely needs more than one search service, ideally on the same layer.  There are different ways to search which are better or worse for different data, different users, and other factors.  Demonstrating them on the same data makes them comparable for people to evaluate and decide which works best for their case.  

For the demo generally, I think that one demo is easiest to maintain and test but agree that it should be configured in a non-conflicting way.  If we have configurations that are divergent enough, I can see how maybe two demos would be a good idea to demonstrate but then we would have to have two clearly defined use cases and reasoning why we configure each one how we do.

theduckylittle:
We could add the single-term to many-field example to the demo with no changes but to the mapbook and app.js.

klassenjs:
That works for me.

PR #480, opened 2020-04-12
Update to ol-mapbox-style

theduckylittle:
Previous Mapbox GL to OL style library was deprecated.
This new version is more full featured. Updated the
vector layer class to also check the spec for divving up
the properties between paint and layout sections.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -215,18 +215,26 @@
       "dev": true
     },
     "@mapbox/mapbox-gl-style-spec": {
-      "version": "9.0.1",
-      "resolved": "https://registry.npmjs.org/@mapbox/mapbox-gl-style-spec/-/mapbox-gl-style-spec-9.0.1.tgz",
-      "integrity": "sha512-j3qgGIk1LDYlDOo7VGEge5o/rHkhW4Z+w9fSdgbq8ZEg8KkXdxlDZeUbeaO14s4n0ML6Nz7hSF8JNlwYXl8R7A==",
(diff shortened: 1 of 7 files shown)


klassenjs:
Are there any backwards compatibility issues here?

theduckylittle:
Not as far as I can tell:
1. I patched them using the paint vs layout detection.
2. All of our internal query functions used the query functions from `util.js` which wraps and normalizes the queries.
3. All of the layers in the demo appear to work with no changes.

theduckylittle:
oh, and 4. It's all implementing version 8 of the MapBox GL spec. So it should all be compatible.

klassenjs:
I'm getting `et.setRotateWithView is not a function` in the production build on FF when drawing in the Drawing and Markup layer (polygon and point tested).

theduckylittle:
Did you wipe node_modules?

klassenjs:
Yep, twice.

theduckylittle:
Updated the package.json to spec OL6. It must've been my own messed up node_modules!

klassenjs:
Identify is broken.  Identify.  Click on a parcel.

Blank map.  0 Features/1 Layer/Spinny arrows.  Errors in console.
`Unhandled promise rejection TypeError: "c.getGetFeatureInfoUrl is not a function"`

theduckylittle:
Identify should now be fixed. Actually was an issue with any point-styled selection.

klassenjs:
Looks good except for the failing test.

PR #483, opened 2020-04-14
Add Labels to Sketches and Attribute Editing

theduckylittle:
1. Adds a new editor reducer, actions, and modal dialog.
2. Added the ability to edit a label for the sketch layer.

Draw a thing
image

Click the new edit properties icon
image

Click on the feature and get a dialog!
image

Now see label!
image

Depends on PR #480


chughes-lincoln:
I tested it out on my build and it's working very well, the label's also retained in the print.

My main feedback is that it's added to the center of the sketch. This is probably what we'd want for 
polygons, but perhaps not for lines and points. I don't know if it would be possible or desirable to 
program in an offset around the sketch? See my (apparently very large) example:

image

There's also the scenario that users will want to label something on their map, rather than label a 
sketch (i.e. directly place a label like in GeoMoose 2). This can essentially be replicated by placing
a point where they want their label and then labeling the point, but it doesn't look very good in my
opinion with the label directly on the point.

PR #485, opened 2020-04-18
Add "Buffer from Select" feature

theduckylittle:
1. Major clean up of selection feature state.
2. Adds a new small "buffer-select" services to the app
3. Adds the call to buffer-select to the desktop demo.

New buffer button to buffer feature
image

And does this
image

### DIFF
--- a/docs/howto/change-buffer-units.md
+++ b/docs/howto/change-buffer-units.md
@@ -1,0 +1,22 @@
+# How-to change the default buffer units
+
+The buffer tool defaults to using feet. This can be changed with the map
+configuration when initializing the application by changing the `defaultUnits`
+settings.
+
+In `app.js`:
+
+```
+var app = new gm3.Application({
+    mapserver_url: CONFIG.mapserver_url,
+    mapfile_root: CONFIG.mapfile_root,
+    map: {
+        scaleLine: {
+            enabled: true,
+            units: 'imperial'
+        },
+        defaultUnits: 'm'
+    }
+});
+
+```
--- a/docs/howto/index.md
+++ b/docs/howto/index.md
@@ -26,4 +26,4 @@ These are intended to be quick helpers for enhancing a GeoMoose application.
 * [How-to style the header](./style-the-header.md)
 * [How-to display a raster layer and query as a vector layer](./query-as-vector.md)
 * [How-to add Attribution](./add-attribution.md)
-
+* [How-to change the default buffer units]('./change-buffer-units.md')
--- a/examples/desktop/app.js
+++ b/examples/desktop/app.js
@@ -149,6 +149,7 @@ app.loadMapbook({url: 'mapbook.xml'}).then(function() {
     });
 
     app.registerService('buffer-select', SelectService, {
+        drawToolsLabel: '',
         tools: {'buffer': true},
     });
 
--- a/examples/desktop/app.js
+++ b/examples/desktop/app.js
@@ -32,14 +32,12 @@
 var app = new gm3.Application({
     mapserver_url: CONFIG.mapserver_url,
     mapfile_root: CONFIG.mapfile_root,
-
     map: {
         scaleLine: {
             enabled: true,
             units: 'imperial'
         }
     }
-
 });
 
 app.uiUpdate = function(ui) {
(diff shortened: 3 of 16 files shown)


klassenjs:
Now this is cool.  Lot's of possibilities to plumb stuff together here.

Bugs:
1) Select -> Buffer Select 1000ft -> Buffer Select (still says 1000ft but is 0ft until you update the "with buffer" input.  I'd guess the units drop down might have the same issue.
2) Identify doesn't auto-go.  When pick point and click Go returns "No Valid Selection Geometry"
2') Select point,multi-point, line returns nothing

klassenjs:
eslint is unhappy.  I'm not sure the correct fix for the second error.  Looks like a hash, but I'm not sure about the JSX syntax.  I think the right thing to do is remove the comma.

```
  274:1   error  Expected indentation of 12 spaces but found 11  indent
  357:46  error  A space is required after ','                   comma-spacing

✖ 2 problems (2 errors, 0 warnings)
  2 errors, 0 warnings potentially fixable with the `--fix` option.
```

The "With Buffer" in the new service is still not right on the map the second time through.  (The input retains the value from the first time, but the map displays buffer=0 until editing the input or changing the units).

Measure tool doesn't display anything in the Super Tab.

Select Features and Identify seem to be working now.

klassenjs:
Strange one in Select Features.  Clicking in the buffer box and backspace to remove the zero locks up the browser.

chughes-lincoln:
Thanks for working on this, it's looking very promising! I haven't had a chance to test it yet but I'll do so soon.

Any thoughts on changing the icon to mapskin's buffer icon or something along those lines? 
The right angle arrow doesn't intuitively feel like a buffer tool to me.
image

klassenjs:
@chughes-lincoln: I second the bit about the icon being non-obvious.

theduckylittle:
New commit handles temporary editing state better and fixes the icon.

klassenjs:
It might actually be not broken.

klassenjs:
Waiting for @chughes-lincoln before merging.

chughes-lincoln:
I have a few comments. With the select service and the demo configured as is, I'm not noticing any issues, although I have the nitpicky comment that the 'Using' parameter is still there but is no longer relevant to the user.

I'm noticing larger issues when you start messing around with your configuration:

1)  You can't remove the Buffer options from the initial select service without also removing them 
from the buffer from select service. It may be desirable for the admin to disable the buffer options
in select and have the users just use the buffer from results (although the buffer in the select service still allows you to buffer sketches). 

2)  You can't turn autoGo on. (You would have to turn the buffer options off in the select service which removes them from buffering from results, and even if you do, I'm getting an error when you click the initial buffer button in your results grid). This may be because it's not compatible with buffering as you need to add an input, but ideally you'd be able to turn it on for the initial selection and off buffering from results.

On the admin side I suppose you could probably fix both of the above by referencing a separate javascript file for the buffer from results and turning the various parameters on or off individually.

3)  When I turn on autoGo, after an initial selection, it reruns the selection any time I click on the select tool rather than waiting for me to choose my options and select features on the map.

theduckylittle:
Re: autoGo. We've kinda gone over this a few times. If there is any user input required, autoGo is not an option. This includes needing to set the buffer distance.

Re: You don't need to change the Javascript file. You need to pass-in "tools: {}" to the Select service options to remove "buffer" from the options for select and buffer select should be unaffected. But you do need to individually specify the tools you want enabled. If you do "tools: {Buffer: false}" it will just show now inputs.

Re: "Using" label. I've updated the select service and the example app to remove the label.

chughes-lincoln:
Is there a way of overwriting autoGo in app.js? I've tried overwriting it in the same way as the title and tools parameters (while setting only the options where autoGo is compatible). To remove the buffer options you also need to set bufferAvailable to false in select.js; my attempts to overwrite it here were unsuccessful. 
```
          app.registerService('select-taxlots', SelectService, {

            title: 'Select Taxlots',
            tools: {
                'Box': true,
                'Point': true,
                'MultiPoint': false,
                'Polygon': true,
                'LineString': true,
                'Select': false,
                'Modify': false,
                'default': 'Box',
                'buffer': false
            },

            bufferAvailable: false,
            autoGo: true,

            fields: [{
              type: 'hidden',
              name: 'layer',
              default: 'Taxlots_selection-wfs/Taxlots_selection'
                    }]
            });
```

theduckylittle:
autoGo is not supported by the Select service. It is set to `false` in the services source code with no support for changing that setting in the options.

As a protocol note, the Pull Request is not an appropriate place for querying about how to improve an individual configuration. Any comments should be germane to whether the PR functions as intended and/or whether its the correct intention to the desired spec.

chughes-lincoln:
Okay, thanks for your feedback.

PR #486, opened 2020-04-23
Fix broken keepAlive setting and add it to Select.

theduckylittle:
1. `keepAlive` will keep the users last draw tool active for the service.
    When the user draws a new feature then it will go back to the service-form.
    This is in contrast to `autoGo` which assumes no input from the user.
 2. The `keepAlive` property is now set to true in the demo app.
    This show cases how the user can repeatedly draw a new polygon.
    And keep changing settings.

### DIFF
--- a/examples/desktop/app.js
+++ b/examples/desktop/app.js
@@ -146,6 +146,7 @@ app.loadMapbook({url: 'mapbook.xml'}).then(function() {
     app.registerService('select', SelectService, {
         // set the default layer
         defaultLayer: 'vector-parcels/parcels',
+        keepAlive: true,
     });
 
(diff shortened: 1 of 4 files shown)


theduckylittle:
@chughes-lincoln - I think this might be closer to the behaviour you'd like to implement for select.

chughes-lincoln:
@theduckylittle Thanks for this feedback! It's definitely helpful to know this can be configured.

klassenjs:
I'm good with this one if you think it is ready.  It makes select act more in line with identify and measure.

theduckylittle:
I think it's good to go!

PR #492, opened 2020-05-14
Support a list-of-ilikes search

theduckylittle:
This builds on the single-search terms search.
It can search multiple fields and the order of terms won't matter.

For example "terry little" and "little terry" will now return the
same list.


chughes-lincoln:
I've tested this out on my build and it's working great. It definitely helps when the data isn't formatted like the user expects it.

PR #493, opened 2020-05-15
Additional Service Options

theduckylittle:
1. Allow a service to zoom to its results after it's finished.
2. Allow the grid to stay minimized when the service finishes (instead of always being "maximized").

### DIFF
--- a/examples/desktop/app.js
+++ b/examples/desktop/app.js
@@ -133,7 +133,8 @@ app.loadMapbook({url: 'mapbook.xml'}).then(function() {
                 validateFieldValuesResult.message = 'Please complete at least one field.'
             }
             return validateFieldValuesResult;
-        }
+        },
+        zoomToResults: true
     });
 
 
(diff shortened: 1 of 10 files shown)


chughes-lincoln:
I tested it out on my build and I have some feedback (a few minor issues):

1. Zoom to results - For the most part it works very well. I noticed that the search service appears
to hang when the search finds no results (an error gets thrown in the dev console and the spinner spins as if it's continuously searching for results). I'm assuming this is because the service is telling it to zoom to results that don't exist. A couple errors also get thrown in the dev console when clearing results (regardless of whether you've configured zoomToResults for that particular service), though it doesn't look like anything unexpected happens that the user can see.

2. Allow grid to start minimized - This is working great!

Thanks for your hard work!

theduckylittle:
Great catch! I've fixed the bug by adding a safety check and then wrote a test to verify the conditions.

PR #494, opened 2020-05-17
Add scale-based smarts to WMS legends

theduckylittle:
1. Adds new scale calculation function (heavily lifted from the OL code base).
2. Adds `SCALE` parameter to WMS GetLegendGraphic requests.

### DIFF
--- a/src/gm3/components/catalog/legend.js
+++ b/src/gm3/components/catalog/legend.js
@@ -43,7 +43,11 @@ class CatalogLegend extends Component {
     }
 
     renderLegend(src) {
-        const legend = getLegend(this.props.mapSources[src.mapSourceName], this.props.mapView, src.layerName);
+        const legend = getLegend(
+            this.props.mapSources[src.mapSourceName],
+            this.props.mapView,
+            src.layerName
+        );
 
         const key = 'legend_' + src.mapSourceName + '_' + src.layerName;
         let legend_idx = 0;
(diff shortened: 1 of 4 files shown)


klassenjs:
This looks like it is using 90.7 dpi (.28mm dot pitch) to calculate scale (as per the WMS 1.3.0 spec §7.2.4.6.9). It looks like CSS uses 96dpi as "standard",  MapServer uses 72 dpi by default for calculating scale.  So at least for a default MapServer WMS, the legend returned won't quite match the map displayed.  I'm not sure what is the most reasonable definition.

Maybe this we should do nothing (and claim we are following the spec, get over it)?  Maybe this is a docs issue?  Maybe we should be setting DEFRESOLUTION in our geomoose_globals.map so MapServer more closely follows the spec?

elil:
Changing DEFRESOLUTION in our geomoose_globals.map so MapServer matches 90.7 (and adding a note in the documentation somewhere that WMS legends expect ~91 dpi resolution) seems like a good approach.  For the purposes of displaying a legend, ~91 may approximately equal 96 dpi.

theduckylittle:
Made a bit of an oopsy and pushed to gm3-demo-data master. Here is the change and the note re: DEFRESOLUTION:
- https://github.com/geomoose/gm3-demo-data/commit/26331b74fdf03a6fb7d5e63b8b3cc002bc1eea9a

klassenjs:
Meh, it's gm3-demo-data.

PR #496, opened 2020-05-19
Add mulitple feature selection + buffer

theduckylittle:
Using the shift-key lets the user select multiple features
from a set of results.

Buffers are also supported.


chughes-lincoln:
I tested it out on my build and it's working very well (both for selecting results and Drawing and Markup)

Issue #495, opened 2020-05-18
End Drawing Text


chughes-lincoln:
It looks like with the additions for language support (See https://github.com/geomoose/gm3/pull/489), some of the text for ending drawings in the select tool didn't make it through (for Draw Box, Draw Multi-Point, and Modify)

![image](https://user-images.githubusercontent.com/43766648/82237725-fa759300-98ea-11ea-9a7b-ec8472d782d7.png)

chughes-lincoln:
Fixed by https://github.com/geomoose/gm3/pull/497

PR #497, opened 2020-05-20
Fix translation of drawing tools.

theduckylittle:
The dictionary was incomplete. This simplifys the logic.

refs: #495


chughes-lincoln:
This fixes the issue for me. Thanks!

PR #499, opened 2020-05-28
Fix missing plugin-doc link.

theduckylittle:


theduckylittle:
Self merging a docs update.

PR #518, opened 2020-06-01
update MS4W quickstart

jmckenna:
- update quickstart with latest requirements + versions

(requested by @brentfraser )


docs/ms4w-quickstart/index.md
```diff
@@ -1,53 +1,91 @@
...
- * A computer running Windows 7 or newer.
- * Download the latest copy of MS4W.  [Download Here.](https://ms4w.com)
- * Download the latest copy of the GeoMoose Demo Data. [Download Here.](https://www.geomoose.org/downloads/gm3-demo-data-3.3.0-ms4w.zip)
- * Download the latest version of the GeoMoose MS4W Package [Download Here.](https://www.geomoose.org/downloads/gm3-examples-3.3.0-ms4w.zip)
+## <a name="req"></a>Requirements
 
-## Install MapServer for Windows
+ * A computer running Windows 7 or newer (Windows 10 is recommended).
+ * Download the latest copy of MS4W [here](https://ms4w.com).
+ * Download the latest copy of the GeoMoose Demo Data (file named `gm3-demo-data`) [here](https://www.geomoose.org/downloads/).
```
klassenjs, line 14: I'm thinking the individual download lines might make more sense under the Option 2 section since they aren't needed for Option 1.  (Or mark them optional.)

Also, I'm thinking it might be best to link to https://www.geomoose.org/download.html so it is easier to find the two files they need.

Thoughts?

jmckenna: Good point about location, I have now moved them into Option2.

Regarding pointing to the exact files, as those links were not maintained I specifically used this wording (such as "Download the latest copy of the GeoMoose Demo Data (file named `gm3-demo-data`)" to avoid users coming to this page a year from now and using old files (for example, I believe the links were from 2018, and several gm3 versions had been released, which meant that recent readers were downloading old versions incorrectly).

jmckenna: Alternatively, we could point to the exact files as you said, but this must be part of each release process.    

Which way do you prefer?  (asking)

jmckenna: Note that this is why I also don't specify an exact MS4W version (that also will quickly be outdated).

jmckenna: @klassenjs by the way, maybe I am misunderstanding your comment, as actually I do point to the download page for those 2 files (see the 2 "here" links at the top of: https://github.com/jmckenna/gm3/blob/ms4w-docs/docs/ms4w-quickstart/index.md#option2

jmckenna: ok, I have conceded and pointed to the exact files.   (now the responsibility is to keep this url uptodate :)

Cheers.

klassenjs: I didn't mean point to exact files.  I liked your change there - keeping a bunch of places up to date at release time is a pain. My suggestion was pointing to the downloads.html page instead of the apache directory index page so that which files they needed (and were current) were more obvious.  The downloads.html page is updated as part of the release process.

jmckenna: ah ok, sorry for my misunderstanding.  Will change that now...

Issue #463, opened 2020-01-28
Using drawer for tools doesn't retain default option


chughes-lincoln:
If you use a drawer to group tools in your toolbar, some of your default options are no longer retained.

To replicate add a drawer around measure, identify, and select (the only tools with a default option - although the default for identify is the only option). Now when you click on identify or select, the default 'Using' parameter isn't checked. It is retained for the measure tool. Potentially related to https://github.com/geomoose/gm3/issues/428 and https://github.com/geomoose/gm3/issues/454 (For the latter, if you click measure and then click select without clicking 'End Drawing', it retains the default from the measure tool in your selection).

theduckylittle:
See PR: #528

chughes-lincoln:
Fixed by #528

PR #527, opened 2020-06-05
Fix for tools starting a service in the drawer

theduckylittle:
The service needed passed into the button.

refs: #463


theduckylittle:
Superseeded with #528

Issue #507, opened 2020-05-28
Remove "tip" from parcel layer in mapbook.xml


brentfraser:
I don't think there are tool tips on catalog layer names on GM3.x?  So likely a cut-and-paste error from a 2x mapbook.  

Although layer name tool tips would be nice...

brentfraser:
And the "legend-toggle" tag too

brentfraser:
Never mind about the "legend-toggle" tag.  I see it is legit.

brentfraser:
Fixed! Nice!

PR #530, opened 2020-06-05
Restore tooltips on layer titles.

theduckylittle:
Like it says on the tin. Uses the `tip` attribute.


theduckylittle:
refs: #507

PR #545, opened 2020-06-15
Examples

brentfraser:
Add a "test" example for testing the various map-source types.


theduckylittle:
I'm not sure we necessarily need the test application? Does this boil down to @klassenjs 's comments about dropping the test grids? There's been a lot of traffic lately but I was fine with simply commenting them out for now.

brentfraser:
It is really just to test the different map-source types, and maybe showcase some of the features.

theduckylittle:
Is there any reason they can't go into demo?

brentfraser:
I've tried to track down public servers to make it easier to test releases.

theduckylittle:
That's great! I'm totally in favor of that. I'm of the mind that the demo is meant to showcase as many features as possible. Obviously, it gets used heavily for testing as well but the fewer demo apps we need to maintain the better. There's already a lot of issues with forgetting to test the mobile app.

brentfraser:
I didn't want to clutter the basic demo which shows well to new users interested in getting an overview.

brentfraser:
Maybe some of the things could go in the basic demo:  
- the layer refresh to the weather radar layer (something that actually changes)
- the more detailed legend examples.

brentfraser:
And maybe move the Grid layer to the test example.

chughes-lincoln:
I don't think it would be too cluttered to add these examples to the demo application, especially if you group them in some way in the catalog (i.e. a group for all of the various ags types).

theduckylittle:
I'm with @chughes-lincoln , I'd actually like to add them to the demo and don't mind a little clutter.

As a concession to clutter, I'm +1 on commenting out the test grid layers. They're still there for "testing" but normal users won't see them. And we can add a note in the comments about what they are for.

brentfraser:
Maybe just put all the test source/layers under a "test" group in the catalog.  The best of both worlds!

brentfraser:
too stale, replace with #671

PR #551, opened 2020-06-22
Add doc for doing a list of likes.

theduckylittle:
New doc to demonstrate how to do an ilike search
on a list of fields given a comma separated list
of input terms.

### DIFF
--- a/docs/howto/configure-search-or-fields.md
+++ b/docs/howto/configure-search-or-fields.md
@@ -130,11 +130,7 @@ Search using the following list: `350010001079, 070030004077, 163030001000, 1600
 
 ## Searching the kitchen sink
 
-This search does the follwoing:
-
- * The user takes input from a single field.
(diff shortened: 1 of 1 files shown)


klassenjs:
@theduckylittle I attempted to cleanup the language some.

The thing I see that may be confusing that I'm not sure is explained anywhere is what the `'search'` and `SearchService` parameters mean to `app.registerService`.  And how the first parameter is linked to the `<toolbar>` in the mapbook, and the implications for creating multiple search tools and avoiding conflicts.

This might be better (or might already exist) in a more generic section about how services work though.

theduckylittle:
https://geomoose.github.io/gm3/actions-and-services.html

This doc could use updating and details. My brain has a lot of shortcuts from being so familiar with the code so someone updating would be great. But I am cool with the changes.

Issue #327, opened 2018-10-02
Printing Enhancements


klassenjs:
1) Print to a specified scale
2) Print legends
3) Print compass rose
4) Print attribution (ex. for OpenStreetMaps)

tchaddad:
+1 for adding back the printing of legends

christian4423:
Maybe add crossOrigin default to Anonymous for printing. 

In OL in image.js the code 

```
  this.image_ = new Image();
  if (crossOrigin !== null) {
    this.image_.crossOrigin = crossOrigin;
  }
```

When printing I get a canvas tainted error. I could  not find in the gm3 code where you all call to ol/image.js to pass it the anon value. 

I changed the ol code to 

```
  this.image_ = new Image();
  if (crossOrigin !== null) {
    this.image_.crossOrigin = crossOrigin;
  }else{
    this.image_.crossOrigin = "Anonymous";
  }
```

for the meantime but it would be nice if gm3 did that out of the box.

PR #552, opened 2020-06-22
Add Legend Printing

theduckylittle:
1. New print layout which demos legends.
2. Fixes issue with needing to expose `setState` of modals
   which is a react anit-pattern (new store-drive `showModal` function
   is available through Applicaiton).
3. Works out all the async business with loading the images
   and then rendering the PDF.

refs: #327


theduckylittle:
Known limitation: This will _only_ work with image-type legends. It will not currently work with the HTML style legends.

PR #558, opened 2020-06-28
Add Experimental API and First Experimental API plugin

theduckylittle:
1. Adds the new `experimental` API namespace to Application.
2. Adds a new basemap-toggle plugin which utilizes those
   new APIs.


klassenjs:
Do we want experimental stuff in the demo?

theduckylittle:
That was actually an accident.

PR #559, opened 2020-06-29
Add new Bookmark Modal

theduckylittle:
We had a similar bookmark "Tab" in the old GeoMoose,
this creates a modal with a textfield that makes it
easy to copy the URL for bookmarks.

### DIFF
--- a/examples/desktop/app.js
+++ b/examples/desktop/app.js
@@ -254,6 +254,13 @@ app.loadMapbook({url: 'mapbook.xml'}).then(function() {
         }
     }, {});
 
+    app.add(gm3.components.BookmarkModal, 'bookmark-modal', {});
+    app.registerAction('bookmark', function() {
+        this.run = function() {
+            app.showModal('bookmark');
+        }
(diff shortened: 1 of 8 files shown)


chughes-lincoln:
I tested it out on my build (after the map.js changes from https://github.com/geomoose/gm3/pull/558 that were needed to fix the build issues). It's working well. My main comment is that if there's a lot of data, the URL can get very long and it can be a bit tedious to copy it:



I modified the bookmark-modal.js file to add a clickable hyperlink between the label and the text box, which might make it a bit less clunky (you can directly right click and copy rather than manually highlighting everything in the text box):

```
            <div>
                <label>This url can be copied and pasted to make a bookmark:</label>
                <a href={'' + document.location} target = "_blank"> Map Link</a>
                <textarea
                    style={{
                        width: '100%',
                        height: '200px',
                        fontFamily: 'mono',
                    }}
                    defaultValue={'' + document.location}
                />
            </div>
```

PR #562, opened 2020-06-29
Fix lint/build errors.

klassenjs:
This should get applied before #558 and #559 then those will likely need to be rebased.

### DIFF
--- a/src/gm3/components/map.js
+++ b/src/gm3/components/map.js
@@ -407,8 +407,6 @@ class Map extends React.Component {
     agsFeatureQuery(queryId, query, queryLayer) {
         // get the map source
         const ms_name = util.getMapSourceName(queryLayer);
-        const layer_name = util.getLayerName(queryLayer);
-
         const map_source = this.props.mapSources[ms_name];
 
         // if the openlayers layer is not on, this fakes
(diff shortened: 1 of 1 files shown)


src/gm3/components/map.js
```diff
@@ -498,7 +496,8 @@ class Map extends React.Component {
             // setup the spatial filter.
             query_params.geometryType = geom_type_lookup[queryGeometry.type];
             query_params.geometry = esri_format.writeGeometry(ol_geom);
-            query_params.spatialRel = 'esriSpatialRelIntersects'; // for lines?:'esriSpatialRelEnvelopeIntersects';
+            query_params.spatialRel = 'esriSpatialRelIntersects';
+            // for lines?:'esriSpatialRelEnvelopeIntersects';
```
theduckylittle, line 500: This should likely go above the query_params line.

klassenjs: Really, it should be implemented or deleted.

klassenjs: It's effectively a TODO: for an unimplemented case.

klassenjs: It doesn't describe what the line above it does, it describes future work.  That's why I put it below.

theduckylittle:
Cool. :+1:

PR #575, opened 2020-07-29
Included clickable hyperlink in bookmark tool

chughes-lincoln:
URL's can be long and larger than the box they're in. A clickable link allows users to right click and copy the link rather than manually copying the entire link (or unintentionally omitting part of the link when copying).


theduckylittle:
Merged using #580 with a clean git history.

Issue #579, opened 2020-08-05
WFS  spatial queries hard-coded to expect "geom" as the geometry name


brentfraser:
Looking at [wfs.js](https://github.com/geomoose/gm3/blob/2e61b05bd9a593a3e04ef0495553fc86ec1ca4b2/src/gm3/components/layers/wfs.js#L74), I see that GeoMoose expects/requires the WFS server layers to have the geometry property named "geom".   Is there a way to over-ride this?

brentfraser:
MapServer doesn't seem to care what the geometry column is called in the query
`"<Intersects><PropertyName>geom</PropertyName>"`
I guess it either ignores it or if the column doesn't exist it just uses the geometry column that does exists.

GeoServer seems to care:
Error
Illegal property name: geom for feature type

brentfraser:
Changing the code to be 
`const geom_field = null;`
seems to work a little better.  OpenLayers sends:
`<Intersects><PropertyName>the_geom</PropertyName>`
which just happens to be the geometry column name of my server's layers.  But I can see that this would not work for all servers.

brentfraser:
Looks good.  But would the param be sent on every request? Or not because it has no "value" attribute?

theduckylittle:
I should double check my syntax for the XML. :laughing: 

It's doesn't get sent because I've also added a new exclusion list for parameters that should not be serialized.

brentfraser:
Here's a server url for testing
https://mrdata.usgs.gov/wfs/sgmc2?service=WFS&version=1.1.0&typeName=ms:Structure&request=getFeature&maxFeatures=1

unfortunately, it only generates geometry in EPSG:4326, so we may need some GM code enhancements to deal with that.

brentfraser:
I've tested with
```
		<config name="geometry-name" value="the_geom"/>
```
and it works well.

PR #584, opened 2020-08-07
Add config param for custom geometry name.

theduckylittle:
Using `<param name='geometry-name' value='columnName' />`
change the name of the column used for WFS queries.

refs: #579


klassenjs:
This seems like it is overloading `<params>` to be more than one thing.  I'm not sure if this is a big deal or not or we are already doing this elsewhere, but this seems like something that could become a problem.

I thought `params` was to pass parameters to the server (possibly overriding a GeoMoose or OL default).  This seems more as a hint for how to interpret the response (maybe more related to `style`?).  I think (although it is likely uncommon) that WFS/GML can return more than one geometry column per feature too.

brentfraser:
Maybe an optional attribute on the layer tag would be a solution:
`<layer name="parcels" geometry-name="myGeom">`

theduckylittle:
It's used on query, not interpretation, which is why I thought `<param>` was appropriate.

It is true that otherwise, `<param>`s are passed along to the service. This is something we have done historically.

@klassenjs - Do you have a better suggestion? I don't like adding another attribute to the `map-source` or the `layer` as that requires more elaborate changes to the parsers, actions, and reducers.

klassenjs:
It seems to fit better on `layer` or `style`.  (You could have multiple layers in a mapsource with different styles that show a different geometry column).

theduckylittle:
I have never actually gotten that to work with MapServer or GeoServer. They both require a single geometry column to be specified...

klassenjs:
Are you happy with it as is or are you planning a change?

theduckylittle:
_I know this is black magic_: What if we added a `wfs:` prefix to the param name? Something that more-or-less clearly indicates it is naming it for WFS? Looking ahead, I see the desire to support other query projections. To do that we'd need more configuration parameters.

I'm also open to adding `<config name="">value</config>`.

theduckylittle:
Latest commit has `<config name= value=>` style. Identical to `<param...>` except this ensures it doesn't need exceptionalized anywhere.

Issue #583, opened 2020-08-07
Set maxZoom for map


brentfraser:
I'd like to set the maximum zoom level of my map.  This is have a better result when:
 - tile servers send bogus tile images for high zoom levels
 - Zoom-to-results on point features zooms in waay too much.

Any thoughts on how to do that?

I've tried:
```
var app = new gm3.Application({
    mapserver_url: CONFIG.mapserver_url,
    mapfile_root: CONFIG.mapfile_root,

    map: {
        maxZoom: 19,
        scaleLine: {
            enabled: true,
            units: 'imperial'
        }
    }

});
```
but no go.

brentfraser:
So in [map.js](https://github.com/geomoose/gm3/blob/2e61b05bd9a593a3e04ef0495553fc86ec1ca4b2/src/gm3/components/map.js#L912) I added:

```
        if(this.props.config.maxZoom) {
            view_params.maxZoom = this.props.config.maxZoom;
        }

        // initialize the map.
```

with
```
var app = new gm3.Application({
    mapserver_url: CONFIG.mapserver_url,
    mapfile_root: CONFIG.mapfile_root,

    map: {
        maxZoom: 19,
        scaleLine: {
            enabled: true,
            units: 'imperial'
        }
    }

});
```
I was able to get the desired behaviour, but it seems like such a hack...

theduckylittle:
It's now

```
....
map: {
  view: {
    maxZoom: 19,
  }
}
...
```

But also allows setting the center, minZoom, maxZoom, and defualt zoom there as well.

brentfraser:
Looks good!

PR #585, opened 2020-08-07
Allow more default zoom settings

theduckylittle:
Adds the ability to set the minZoom, maxZoom, zoom,
and center from the application configuration.

refs: #583

### DIFF
--- a/src/gm3/components/map.js
+++ b/src/gm3/components/map.js
@@ -921,6 +921,15 @@ class Map extends React.Component {
             }
         }
 
+        if (this.props.config.view) {
+            const mixinKeys = ['center', 'zoom', 'maxZoom', 'minZoom'];
+            mixinKeys.forEach(key => {
+                if (this.props.config.view[key]) {
+                    view_params[key] = this.props.config.view[key];
+                }
+            });
+        }
+
         // initialize the map.
         this.map = new olMap({
             target: this.mapDiv,


klassenjs:
This needs some docs.

klassenjs:
Does this replace the `app.setView()` in `app.js` (in terms of preferred method going forward, not in terms of breakage)?

brentfraser:
My perspective:
- min/max Zoom is related to the map definition
- zoom is a current setting, changeable by the user

To some extent we're at the mercy of OL (and I don't know what limitations they impose).  Conceptually there should be no reason why you couldn't [re]set the min/max zoom of the map with a call to setView().

klassenjs:
That's not what I meant.  I meant for how we setup the default view in `app.js` in the examples.  The new parameters (`zoom` and `center`) appear to fill the same roll as the current `app.setView()` call a few lines later (and is more declarative which seems preferred).  I'm not talking about removing the ability to call `app.setView()`.

brentfraser:
Ah.  Depends on the purpose of the examples.  If we want to illustrate the use of some of the available methods, I suppose we could include a view object in the map object def in app.js, then show how the implementer can set the view's center + zoom to something else.

theduckylittle:
Add docs.

No, we don't need to deprecate or prefer the other version. The advantage to using `setView` is you could use information from the mapbook to set the view. Where as doing it with the application's configuration doesn't have access to that.

klassenjs:
The code added `'center', 'zoom', 'maxZoom', 'minZoom'` but documented `minZoom, maxZoom, extent`. Was this intentional?

theduckylittle:
Somewhat, I didn't want to confuse this with `setView`.

Issue #369, opened 2019-04-09
Display grid results after query


chughes-lincoln:
Is there the ability to easily display the grid results when doing a search? 

I was able to return grid information for a search by doing any selection (which creates a table), and then doing a search (which updated the table with the query information), but I wasn't able to simply search and return a grid without doing an unrelated selection first.

The grid provides additional functionality (sorting, filtering, saving results as a csv) that would be useful to be able to do with searches.

klassenjs:
Cool, I didn't know it worked with searches at all.  It was primarily
designed to work with the select tool.

I could see that being useful though.

chughes-lincoln:
I made some minor modifications to the search.js file in the services folder and got the grid to display for searches just like it does if you have it configured to display for selections. Main caveat is that this references the same grid templates as the selection, so it forces you to use a grid for both select and search or for neither, rather than one or the other.

See changes:

https://github.com/geomoose/gm3/blob/master/src/services/search.js#L40 

Added this line of code from select.js after line 40 of the original

`    this.showGrid = options.showGrid !== undefined ? options.showGrid : true;`

Added this section from select.js before Application.dispatchQuery (Line 107)


        
            var templates = [this.template];
            if(this.showGrid) {
                templates.push('@gridColumns');
                templates.push('@gridRow');
            }

And then changed this.template to templates in the Application.dispatchQuery section below

      Application.dispatchQuery(
            this.name,
            selection,
            this.prepareFields(fields),
            this.getSearchLayers(this.searchLayers, fields),
            templates
        );

chughes-lincoln:
Is there any desire to add my changes to show the grid in search results in the demo? Currently search results only return tab results, which limits you a bit in what you can do with them (mainly being able to export results to a csv)

brentfraser:
Yes.  I think this would be a great addition

chughes-lincoln:
I also think this pairs well with the ability to have the grid be minimized by default. Therefore you get your normal tab results, and have the ability to see table results or export to csv if you want without cluttering your map.

brentfraser:
@chughes-lincoln : another solution to "minimized by default" would be to tweak the results tab a bit:

1. Move the big zoom-to-results "button" to a smaller icon on the results tool bar (along with the clear-results icon) 
2. Add a show/hide-table icon to the toolbar for showing the results in the table.

chughes-lincoln:
Thanks Brent, that looks like it could be an effective solution.

[...]

PR #587, opened 2020-08-11
Add grid for identify and search

theduckylittle:
It's optional, see the docs for how to add!

refs: #369


brentfraser:
I see the doc in the commit, but not the code.

theduckylittle:
try now

brentfraser:
Since a single Identify can return results from multiple layers, how would the table look?  Does it render multiple headers?

theduckylittle:
Not sure, that's left as an exercise for the administrator.

brentfraser:
OK thanks!

brentfraser:
So the new template names for Search and Identify are *-grid-columns, but the one for select is gridColumns?

brentfraser:
The Identify on multiple layers results in only one layer's results rendered in the table:
image

elil:
I'm not sure that I understand the concept of a grid for results from multiple datasources (like identify if configured with more than one layer).  A grid makes perfect sense for results from a single datesource/layer (like select, search, or identify if configured with just one layer).

brentfraser:
@eli The original issue was just to add the grid to search results.  But I like that the PR covers identify too (for completeness) even though it is an edge case.

theduckylittle:
I know the first comment after adding search would be someone asking for Identify. "So long as I was in there" I did the work.

I've also added support for `select-grid-columns` and `select-grid-row` templates. `gridColumns` and `gridRow` are still supported for backwards compatibility.

brentfraser:
Nice!

brentfraser:
So it looks like I still need a 
```
            <template name="identify" src="./templates/parcels_points.html" />
```
to get the Identify to execute even though I have
```
            <template name="identify-grid-columns" src="./templates/parcel-columns.json" />
            <template name="identify-grid-row"     src="./templates/parcel-row.html" />
```

I think the logic needs to be "either (or both)" to get results.  The work-around is to have no template for identify:
```
            <template name="identify"/>
```

brentfraser:
This is true for Select as well.  It requires a 
`<template name="select"/>`
or
`<template name="select-header"/>`
for the Select to work even if select-grid-* templates are specified

theduckylittle:
@brentfraser  - Try this latest change. It should no longer require an identify or select template.

brentfraser:
using master + PR code , doing an Identify on Parcels says "No layers to identify!"
Parcel layer is on (all others are off), and:
```
        <layer name="parcels" status="on" query-as="vector-parcels/parcels">
            <template name="identify" src="./templates/parcels.html" />
                    </layer>
        <layer name="parcels_points"/>
```

theduckylittle:
That's not how you config the templates. It's done in the map-source.

brentfraser:
Sorry, that was a snippet of the mapsource (in the desktop demo):
```
    <map-source name="parcels" type="mapserver" up="true" down="true" title="Parcels">
        <file>./demo/parcels/parcels.map</file>
        <layer name="parcels" status="on" query-as="vector-parcels/parcels">
            <template name="identify" src="./templates/parcels.html" />
                    </layer>
        <layer name="parcels_points"/>
        <param name="FORMAT" value="image/png"/>
        <!-- this is used only for testing the group-functionality -->
        <!--
        -->
        <layer name="parcels_group"/>
        <layer name="all"/>
    </map-source>
```

theduckylittle:
I'm not seeing the same bug locally. I'm running in the branch and it appears to be returning identify just fine.

brentfraser:
I'm baffled.  I've done a clone with just the #587 PR:
```
$ git log --oneline --decorate --graph --all
* 7916e42 (HEAD -> master) Do not require 'identify' or 'select' templates for grid rendering
* 13b38f2 Add `select-grid-columns` for consistency.
| * 362709e (refs/stash) WIP on master: d942070 Add grid for identify and search
|/|
| * a270a2b index on master: d942070 Add grid for identify and search
|/
* d942070 Add grid for identify and search
* 8d81393 (origin/master, origin/HEAD) allow passing user params in ags-vector map-source (#582)
* 1aaa451 add boundedby property to WFS vectors (#581)
* 6f3202b Bump elliptic from 6.5.2 to 6.5.3 (#578)
* 0f03be5 Included clickable link (#580)
*   2e61b05 Merge branch 'feature/bookmark-modal'
|\
| * b492415 Add new Bookmark Modal
|/
*   e108ac7 Merge remote-tracking branch 'theduckylittle/expr/layer-toggle-plugin'
```

theduckylittle:
I had actually spoke too soon, I'm on _very_ rural internet right now trying to fix things. I should have a fix for it momentarily.

theduckylittle:
Okay! I found the following:
1. I think there was a bug with identify before. The `query-as` looks for the templates to be in the queryable layer. This, I think, is actually a _bug fix_ instead of a regression. I've updated the mapbook to reflect this.
2. The template array seemed to be working as an *and* condition instead of an *or* condition. This was only a bug for identify. So much for ambition and simplicity.
3. The new query loop is actually a little faster now that I've cleaned up the logic.

brentfraser:
In identify.js, it looks like the list of template names has the "@" decorator, but the `match_fn ` function in mapSource.js expects the names not to have the decorator when it does a lookup in the layer's templates:
```
               const tpl_name = template_names[x];
                template_filter_pass = (templates && typeof templates[tpl_name] !== 'undefined');
```
where
`tpl_name = "@identify"`
and
```
templates: Object
identify: {type: "remote", src: "./templates/parcels.html"}

      "}

```

theduckylittle:
That isn't a bug. The "@..." template naming tells geomoose that the template is in the list for the layer.

brentfraser:
Yeah, I didn't think the bug was there, just the logic didn't make sense to me.  Using the latest changes, the Identify works!  Now on to testing the new templates.  Thanks!

klassenjs:
Is this PR now stable and ready for testing and merging?

brentfraser:
Identify Operation test:

wfs (vector-parcels):
 - identify template:   **pass**
 - identify-grid-columns,identify-grid-row templates:    **pass**
 - no template(s):    **fail**:   layer still searched, feature found and highlighted (bad, I expected "No layers to Identify")
                                       but no attributes displayed (good)

brentfraser:
Select Operation test:

wfs (vector-parcels):
 -  Select template: **fail** (no controls displayed in Super Tab)

brentfraser:
It fails in[ select.js](https://github.com/geomoose/gm3/blob/498ff696caff90af370b3605876c7f93a9a93972/src/gm3/components/serviceInputs/select.js#L44) 

`            this.onChange({target: {value: options[0].value}});`

since options is empty.

theduckylittle:
It's because it's being selected wrong. I'm still working on this fix.

theduckylittle:
Ha! Turns out the `@` was being carried through!!! It just didn't caught because the `@`-naming is used in `getTemplate` and not in the `match_fn`. Red letter day on this bug for me.

Then there was a looping bug with the grid.

Then there was a filter logic bug breaking select.

Small change -> big bugs -> fixing more.

brentfraser:
I did identify, search, and select ops with (and without) grid templates on the "vector-parcels" mapsource and everything worked as expected.  Thanks!  That's a wrap!

PR #612, opened 2020-11-27
Return docs to rst

theduckylittle:


klassenjs:
I'm merging this since the old Markdown docs are no longer working.

Issue #613, opened 2020-12-11
Adding it.json


flaviorigolon:
I'd like to share the it.json file for italian language translation which I attach here; surely it can be improved but it could be a start point.
[it.zip](https://github.com/geomoose/gm3/files/5677989/it.zip)

elil:
Great, thanks for getting the Italian language translation started!  

Are you contributing this under the license here, https://github.com/geomoose/gm3/blob/master/AUTHORS?

flaviorigolon:
Oh yes! I' ve read it and I surely agree.
Thanks to you.

Flavio

Il Ven 11 Dic 2020 13:07 Eli L Adam <notifications@github.com> ha scritto:

theduckylittle:
All merged in! Thanks everyone!

PR #615, opened 2020-12-11
Add initial Italian language translation

elil:
Add initial Italian language translation by @flaviorigolon https://github.com/geomoose/gm3/issues/613


elil:
Add initial Italian language translation

Issue #601, opened 2020-08-26
Identify on WFS point and line layers not returning results


brentfraser:
This may be a similar problem to #392.  Identify makes a GetFeature request with a filter of

```
		<Filter xmlns="http://www.opengis.net/ogc">
			<Intersects>
				<PropertyName>the_geom</PropertyName>
				<Point xmlns="http://www.opengis.net/gml">
					<pos srsDimension="2">-8973326. 953823</pos>
				</Point>
			</Intersects>
		</Filter>
```

I don't know if intersecting a point with a point (or a line) is a valid spatial operation.  We may need to construct a box around the query point and sent that with the request.

brentfraser:
Spatial Query Operations (Identify, Select)
---
| Protocol | Feature Geometry | Query By <br> Point (_Identify_) | Query By <br> Line | Query By <br> Polygon | 
|----------| ---------------- |-------------------- |------------------- |---------------------- |
| WFS      | Point   | **No: Bug** | No: N/A? | Yes |
|          | Line    | **No: Bug** | Yes  | Yes |       
|          | Polygon | Yes     | Yes  | Yes |
| AGS vector | Point | Yes (BBOX #539) | No: N/A? | Yes |
|          | Line    | Yes (BBOX #539) | Yes  | Yes |
|          | Polygon | Yes     | Yes  | Yes |
| Vector file | Point | ?  |  ?   |  ?  |
|          | Line    | ?  |  ?   |  ?  |
|          | Polygon |  Yes  | **No: Bug** | **No: Bug** |
| Local    | Point   | **No: Bug** | **No: Tab Bug** | **No: Tab Bug** |
|          | Line    | **No: Bug** | **No: Tab Bug** | **No: Tab Bug** |
|          | Polygon |  Yes    | **No: Tab Bug** | **No: Tab Bug** |

brentfraser:
The fix for the WFS problem is similar to the AGS fix.  In components\map.js, just after line [326](https://github.com/geomoose/gm3/blob/1e0978c6f8e8a4337da367c094d22d25c91d26c8/src/gm3/components/map.js#L324) insert code to make a small box around the point query geom:

```
        if (query.selection.length > 0) {
            const queryGeometry = query.selection[0].geometry;
            // Since WFS servers don't "intersect" a point with anything, we make a box
            // Make a 4 pixel (2 pixels each way) box around the point (and duplicate 
            //   the first point at the end)
            if (queryGeometry.type === 'Point'){
[...]

Issue #536, opened 2020-06-05
fix for Search of AGS layers


brentfraser:
Not sure about query_params part of my fix:
```
diff --git a/src/gm3/components/map.js b/src/gm3/components/map.js
index e3f98cb..396a95a 100644
--- a/src/gm3/components/map.js
+++ b/src/gm3/components/map.js
@@ -438,7 +438,7 @@ class Map extends React.Component {
                 return name + ' like ' + fix_like(value);
             },
             'ilike': function(name, value) {
-                return name + ' like upper(' + fix_like + ')'
+                return name + " like upper('" + fix_like(value) + "')";
             },
             'eq': function(name, value) {
                 return simple_op('=', name, value);
@@ -485,7 +485,7 @@ class Map extends React.Component {
             // setup the spatial filter.
             query_params.geometryType = geom_type_lookup[queryGeometry.type];
             query_params.geometry = esri_format.writeGeometry(ol_geom);
-            query_params.spatialRel = 'esriSpatialRelIntersects';
+            query_params.spatialRel = 'esriSpatialRelEnvelopeIntersects';//'esriSpatialRelIntersects';
         }

         // build the filter fields.
@@ -495,11 +495,13 @@ class Map extends React.Component {
         }

         const where_str = where_statements.join(' and ');
-
+/*
         query_params.layers = JSON.stringify([{
             layerId: layer_name,
             'where': where_str,
         }]);
+*/
+        query_params.where = where_str;

         // get the query service url.
         const query_url = map_source.urls[0] + '/query/';
@@ -574,8 +576,8 @@ class Map extends React.Component {
         const result_features = [];

         const selection = query.selection;
-        if(selection && selection.geometry && selection.geometry.type === 'Point') {
-            const coords = selection.geometry.coordinates;
+        if(selection && selection[0].geometry && selection[0].geometry.type === 'Point') {
+            const coords = selection[0].geometry.coordinates;
             src.forEachFeatureAtCoordinateDirect(coords, (feature) => {
                 result_features.push(format.writeFeatureObject(feature));
             });
```

testing with 

```
    <map-source name="runways" type="ags-vector">
        <url>https://gis2.co.dakota.mn.us/arcgis/rest/services/DCGIS_OL_Transportation/MapServer/21</url>
        <layer name="runways" selectable="true" title="Runways">
            <template name="identify" auto="true" />
            <style><![CDATA[
            {
                "line-color" : "#010138",
                "line-width" : 2
            }
            ]]></style>
[...]

Issue #392, opened 2019-05-13
Identify on Demo's AGS Dakota County Rail layer not returning results


brentfraser:
Go to GeoMoose Demo site
Turn on AGS Dakota County Rail layer
Click Identify button 
Click on rail feature in map.

No results are shown

brentfraser:
I did some investigation using ERSI's web interface to do a point query::

http://gis2.co.dakota.mn.us/arcgis/rest/services/DCGIS_OL_Transportation/MapServer/16/query?where=&text=&objectIds=&time=&geometry=-10389315.44%2C5584710.51&geometryType=esriGeometryPoint&inSR=102100&spatialRel=esriSpatialRelIntersects&relationParam=&outFields=*&returnGeometry=true&returnTrueCurves=false&maxAllowableOffset=&geometryPrecision=&outSR=102100&returnIdsOnly=false&returnCountOnly=false&orderByFields=&groupByFieldsForStatistics=&outStatistics=&returnZ=false&returnM=false&gdbVersion=&returnDistinctValues=false&resultOffset=&resultRecordCount=&f=html

And that did return a record.  Shaving off the last digit from from the point's x and y coord resulted in no records found.  Yikes!

So we may need to include a radius/buffer distance for the point in our query i/f and submit an envelope geom instead of point geom.  ArcGIS 10 may have more/better options...

theduckylittle:
Is there an extra parameter we should be passing in by default?

brentfraser:
Not that I know of (but I haven't done intense Googling).  I didn't find any solid examples of querying by point.

brentfraser:
I think we may need to construct a small bbox, and use that in the query,  Intersecting with a point could be problematic; intersecting with a small rectangle (size based on zoom/scale) would be more predictable.

brentfraser:
In map.js, around line 488 change 
`query_params.spatialRel = 'esriSpatialRelIntersects';`
to: 
`query_params.spatialRel = 'esriSpatialRelEnvelopeIntersects';`
and the Identify works.

At least it works with query geom= point  and feature geom = line.  Not sure about other combinations.

theduckylittle:
So the real answer is that there is a separate Identify function for AGS, however, the server we are using in the demo _does not_ support that function:

- https://developers.arcgis.com/rest/services-reference/identify-map-service-.htm
- https://gis2.co.dakota.mn.us/arcgis/rest/services/DCGIS_OL_Transportation/MapServer/16/

[...]

PR #620, opened 2020-12-29
Adding WFS-T editing to GeoMoose

theduckylittle:
Big honking MR to add WFS-T editing capabilities to GeoMoose.


src/gm3/components/map/index.js
```diff
@@ -1062,67 +1007,110 @@ class Map extends React.Component {
...
-                            }
+                    this.drawTool.on('drawend', evt => {
+                        const querySourceName = util.getMapSourceName(layer.queryAs[0]);
+                        const querySource = this.props.mapSources[
+                            querySourceName
+                        ];
+                        let queryFeature = util.featureToJson(evt.feature);
+                        const mapProjection = this.map.getView().getProjection();
+
+                        if (querySource.config['buffer-point']) {
```
theduckylittle, line 1077: @klassenjs and @brentfraser - Thoughts on this pattern for querying points/lines layers?

brentfraser: Related to #392, #536, #601.  It's an OK fix.  But someone should do some research on the OGC protocols to see if there is something in the protocol to do a spatial query byt point geom.

brentfraser:
Issue:  Delete -> click on empty area (e.g. no polygon) shows "Remove ..." dialog.  No dialog should be shown.

brentfraser:
Issue:  Should the edit (and other?) tools be hidden when layer is turned off/disabled/not shown?

brentfraser:
**Issue:**  If the editable layer is accessed through the "Favourites" tab, no "Edit feature properties" dialog is shown.

brentfraser:
**Issue:** Map tools (zoom in/out, end drawing buttons) obscure menu drop down choices.  Change menu text from "Zoom to Full Extent" to "Full Extent",  This move the other menu items slightly to the left.  Start a feature edit, then click search and observe how the menu options are obscured by the Map tools.

image

brentfraser:
**Issue:** When the WFS-T server is TinyOWS, and an attempt is made to edit properties on a features property values when they are all null, the fetch in wfs.js:171 fails:
```
               jsonFeature.properties.boundedBy = feature.getGeometry().getExtent();
```
because ` jsonFeature.properties` is `null` (a little unexpected).

The fix is to test for null, and make the value an object:

```
                  if (jsonFeature.properties == null) {
                      jsonFeature.properties = [];
                  }
                  jsonFeature.properties.boundedBy = feature.getGeometry().getExtent();
```

or something similar.

theduckylittle:
> Issue: Delete -> click on empty area (e.g. no polygon) shows "Remove ..." dialog (and clicking Okay shows error in console). No dialog should be shown.

Fixed. Wrapped the entire block in the feature-length check.

>  Should the edit (and other?) tools be hidden when layer is turned off/disabled/not shown? Currently the legend is not shown in that case.

This really isn't an "issue" we could hide the tools. My initial thought is to keep them visible since they indicate the capabilities of the layer before it's turned on.

> Issue: If the editable layer is accessed through the "Favourites" tab, no "Edit feature properties" dialog is shown.

Moved the Modals' definition to the map from the catalogs.

> Map tools (zoom in/out, end drawing buttons) obscure menu drop down choices. Change menu text from "Zoom to Full Extent" to "Full Extent", This move the other menu items slightly to the left. Start a feature edit, then click search and observe how the menu options are obscured by the Map tools.

- Added some more Z to the drawers in the toolbar.
- Renamed the Zoom to Full Extent tool.

> `boundedBy`

Added a safer mixin to prevent errors.

brentfraser:
>>    Should the edit (and other?) tools be hidden when layer is turned off/disabled/not shown? Currently the legend is not shown in that case.

> This really isn't an "issue" we could hide the tools. My initial thought is to keep them visible since they indicate the capabilities of the layer before it's turned on.

Currently I put the editable layers in a separate group to indicate which are editable since that ability is so important to the user.

brentfraser:
@theduckylittle The latest commits are awesome!  I've tested my "Issues" and all the bugs are fixed. As for the issue/opinion of "hide edit tools", having a clean looking catalog/layer list is important (as is having an understandable UI).  One solution is to have  an edit icon at the same level of the "refresh" and "metadata" icons on the layer name:

image

and while we're at it, move the "Toggle legend visibility" icon up there too.

theduckylittle:
> @theduckylittle The latest commits are awesome! I've tested my "Issues" and all the bugs are fixed. As for the issue/opinion of "hide edit tools", having a clean looking catalog/layer list is important (as is having an understandable UI). One solution is to have an edit icon at the same level of the "refresh" and "metadata" icons on the layer name:
> 
> image
> 
> and while we're at it, move the "Toggle legend visibility" icon up there too.

I'm willing to open up the discussion on this. There are good/bad reasons for this and I'm willing to talk about them but not in this already monster sized pull request.

brentfraser:
@theduckylittle A discussion on the layer controls would be great!

theduckylittle:
@klassenjs - I believe this is ready for code review.

@brentfraser has done a very thorough job of testing the functionality. I suspect any more bugs will be corner cases are minor oversights that can be handled in a follow up PR.

Commits can/will be squashed appropriately after review.

I suspect it's okay for docs to come in as a separate PR. I think Brent has those in the works.

examples/desktop/app.js
```diff
@@ -49,7 +49,7 @@ app.uiUpdate = function(ui) {
     }
 }
 
-app.loadMapbook({url: 'mapbook.xml'}).then(function() {
+app.loadMapbook({url: 'mapbook-editing.xml'}).then(function() {
```
klassenjs, line 52: This should be reverted before merge and another example created (or something)

tests/gm3/reducers/editor.test.js
```diff
@@ -39,16 +39,20 @@ describe('test the `edit` reducer', () => {
 
         let nextState = reducer();
 
-        nextState = reducer(nextState, setEditFeature(fakeSource, fakeFeature));
+        nextState = reducer(nextState, setEditFeature(fakeFeature));
```
klassenjs, line 42: Kill fake-source on line 38

klassenjs (changes requested):
Also, it might be worth bumping the version in package.json before the merge.  No way this is a precursor to 3.7.1.  Not sure if we want to call this 3.8 or 4.0.

klassenjs:
Drawing and markup->Draw Point
Tries to scroll map instead of adding a point.

klassenjs:
The CSS changes are mostly subtle but it looks way nicer and feels more intuitive than before.

klassenjs:
I'm still working on a Geoserver install.

examples/desktop/app.js
theduckylittle: Moved to a different mapbook, removed from the default app.js

klassenjs (approved):

PR #622, opened 2021-01-16
New json file, first version to Brazilian Portuguese.

EngenheiroUchoa:
I forward a first version of json to Brazilian Portuguese. Brazilians often prefer to keep terms in English rather than trying to translate. So it is possible that we will have debates on the best way to translate some terms. I will help in this process with the Brazilians.


theduckylittle:
This is great! But @EngenheiroUchoa can you add yourself to the AUTHORS file? Adding your name to the AUTHORS file means you agree that these contributions are provided free and clear to the GeoMoose project.

Thanks again!

EngenheiroUchoa:
The requested modification has been made.

PR #623, opened 2021-01-25
Add USNG Map Collar Support

theduckylittle:
Adds a new "usng" layer type for rendering USNG map collars

### DIFF
--- a/examples/desktop/mapbook.xml
+++ b/examples/desktop/mapbook.xml
@@ -4,6 +4,10 @@
         The mapping services define the source of the mapping data.
     -->
 
+    <map-source name="usng-grid" type="usng">
+        <layer name="usng-grid" status="off" />
+    </map-source>
+
     <map-source name="sketch" type="vector" title="Drawing and Markup">
         <layer name="default" selectable="true" status="on">
             <style><![CDATA[
@@ -493,7 +497,8 @@
             <layer title="Weather Radar" src="iastate/nexrad-n0r" />
         </group>
 
-        <group title="Test Grids">
+        <group title="Grids">
+            <layer src="usng-grid/usng-grid" title="USNG Grid" />
             <layer title="1 km grid" src="grids/grid_1km"/>
             <layer title="1 mile grid" src="grids/grid_1mile"/>
         </group>
--- a/package.json
+++ b/package.json
@@ -31,13 +31,14 @@
       "react"
     ],
     "plugins": [
-      "syntax-dynamic-import"
+      "syntax-dynamic-import",
+      "transform-object-rest-spread"
     ]
   },
   "jest": {
     "testURL": "http://localhost",
     "transformIgnorePatterns": [
-      "node_modules/(?!ol|jsts)"
+      "node_modules/(?!ol|jsts|usng)"
     ],
     "moduleDirectories": [
       "./src",
@@ -64,6 +65,7 @@
     "babel-eslint": "^8.2.6",
     "babel-loader": "^7.1.4",
     "babel-plugin-syntax-dynamic-import": "^6.18.0",
+    "babel-plugin-transform-object-rest-spread": "^6.26.0",
     "babel-preset-env": "^1.6.1",
     "babel-preset-es2015": "^6.24.1",
     "babel-preset-react": "^6.24.1",
@@ -108,5 +110,8 @@
     "webpack-cli": "^3.3.10",
     "webpack-dev-server": "^3.10.1",
     "whatwg-fetch": "^3.0.0"
+  },
+  "dependencies": {
+    "usng-map-collar": "^1.0.1"
   }
 }
--- a/src/gm3/components/layers/usng.js
+++ b/src/gm3/components/layers/usng.js
@@ -1,0 +1,35 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2016-2021 Dan "Ducky" Little
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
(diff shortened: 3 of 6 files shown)


package.json
```diff
@@ -31,7 +31,8 @@
       "react"
     ],
     "plugins": [
-      "syntax-dynamic-import"
+      "syntax-dynamic-import",
+      "transform-object-rest-spread"
```
theduckylittle, line 35: This is to add support for the spread syntax ( `{... stuff}`) which is used by the map collar library.

src/gm3/components/layers/usng.js
```diff
@@ -0,0 +1,35 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+import UsngGrid from 'usng-map-collar/UsngGrid';
+
+export const createLayer = () => {
```
theduckylittle, line 27: Double arrow ( `=>`) unctions are context less, aka no `this`, which when supported means they use less resources. I believe these are just transpiled to standard functions for now.

src/gm3/components/map.js
```diff
@@ -703,25 +709,6 @@ class Map extends React.Component {
         }
     }
 
-    sortOlLayers() {
```
theduckylittle, line 706: This is just churn removal.

webpack.config.js
```diff
@@ -71,7 +71,7 @@ module.exports = env => {
                     path.join(__dirname, 'node_modules/'),
                 ],
                 exclude: function(absPath) {
-                    var acceptable = ['ol', 'mapbox-to-ol-style', '@mapbox', 'jsts'];
+                    var acceptable = ['ol', 'mapbox-to-ol-style', '@mapbox', 'jsts', 'usng-map-collar'];
```
theduckylittle, line 74: Let's webpack know that `usng-map-collar` should be transpiled.

examples/desktop/mapbook.xml
```diff
@@ -4,6 +4,10 @@
         The mapping services define the source of the mapping data.
     -->
 
+    <map-source name="usng-grid" type="usng">
+        <layer name="usng-grid" status="on" />
```
theduckylittle, line 8: It may make sense to turn this off before merging as USNG will not really be globally applicable.

brentfraser:
Map collars.  Interesting!  Was there an Issue filed related to this PR?  Is there a description of added capabilities?

theduckylittle:
It adds the USNG map collar? I didn't have an issue for this but it was a request I received on a backchannel. Seemed like it would be a reasonable enough PR.  I can draft a ticket with a story and acceptance criteria.

brentfraser:
I'm just curious if it means what I think it means: a margin surrounding the mapped area containing grid/graticule coordinate values.  In the interactive map?  Just the printed map?

theduckylittle:
It's in the interactive map. You can see an example outside of GeoMoose here: 
- https://theduckylittle.github.io/usng-map-collar/

theduckylittle:
Oh, and also in the printed map.

brentfraser:
Ok.  Years ago I was doing similar things with Mapserver
alberta_v5 0 2_postfix_norot

brentfraser:
So I was thinking of white margins, lat/lon, N/E labels, tick marks, etc.

brentfraser:
All towards the "holy grail" of map rendering margin information: 

image

Not that I have a specific need; I just like good looking maps.

klassenjs:
The lines go wonky south of the equator.

image

Just panning south a little from the above.

image

klassenjs:
Otherwise, generally seems to work.

theduckylittle:
That's a bug in the underlying library.

We can leave this off for now but I know there is interest in the feature.

klassenjs:
It might just be too early in the morning for me, but what bug is in what library?

theduckylittle:
The navigating south and causing the lines to go errant is a bug in the
usng-map-collar library.

klassenjs:
+1 to merge from me.  

Personally, I'd leave the USNG layer enabled in the example mapbook if it doesn't cause problems.  I like the idea of showcasing what is available, and it also makes regression testing easier when the features are enabled.  Off by default would be fine.

brentfraser:
Might be nice to add Lat/Lon and XY options too.

klassenjs:
@brentfraser I agree, that lat/lon and xy grids would be useful to many people, but I don't think that needs to be in this PR.

USNG/MGRS ends up being a special case of grids.  It has a lot of rules around labeling (which are primarily meant for paper maps and which we technically only partially meet, but this is usable and a lot better than nothing).  It also has a lot of gotchas in implementation because it is a "grid" in multiple projections at once (all UTM zones and PS at the poles).  Most other grids are much simpler to implement (and this code could provide a template for doing so).  MGRS also has some irregular grids, which we currently ignore, around Norway where some UTM zones were extended past their natural boundaries.

Personally, I'd like to add the Maidenhead grid as a grid and for coordinate display.  It is just specially labeled 2x1 degree lon/lat boxes and subdivisions.  But that is also a separate issue from this PR.

brentfraser:
@klassenjs  No problem, my comment was more for selecting good keywords.  I guess we could add  `<map-source name="geographic-grid" type="geographic">` and `<map-source name="xy-grid" type="xy">` etc.  in keeping with the current implementation.  Then there would be the need to specify the coordinate system.

theduckylittle:
I'm all for merging and ticking up the lat/long grid. That's built into openlayers and a really simple add if someone wants to pick it up.

Issue #304, opened 2018-05-23
Coordinate Precision


klassenjs:
This started out as just a Lat/Lon issue, but looks like it may apply more broadly.  

Lat/Lon: The current three decimal places is only good to somewhat over 100m (at 45°N).  There may or may not be reasonably room for more digits in cursor tracking area, but we should probably show more digits in measure point.  Ideal would be to show the number of digits appropriate to the current zoom level and latitude.

klassenjs:
(Similar scaling number of digits for X,Y might be good too).

klassenjs:
Also applies to all forms of the measure tool.

tchaddad:
I think this item is closable, BUT I've received a request to show 6 decimals in the lat / lon footer, if that is not too difficult of an amendment to add here...

klassenjs:
It's not closable yet.  Only USNG got fixed.  Lat/Lon and X/Y should also be fixed to show an appropriate # of digits based on zoom level.

tchaddad:
I'd like to prompt on this ticket in case it is doable in the next release. Just got a request about it today....

Specifically the issue of displaying 6 decimal places in the lat / lon footer.

If I should break that out into a separate ticket I can do that.

klassenjs:
This ticket is more about displaying the "right" amount of precision based on zoom level.  

If you want 6 decimal places (always), I think that can already be done in app.js (for latlon and xy types):

```
    var point_projections = [
        {
            label: 'X,Y',
            ref: 'xy',
            precision: 0
        },
        {
            label: 'USNG',
            ref: 'usng'
        },
        {
            label: 'Lat,Lon',
            ref: 'latlon',
            precision: 6
        }
    ];
```

tchaddad:
Ah nice - thanks for pointing to that!

PR #626, opened 2021-02-07
Add Automatic Precision to Lat/Lon Display

theduckylittle:
As the zoom level changes, the user will be presented with
different levels of precision on the latitude and longitude.

refs: #304


klassenjs:
Does this preserve the user being able to set the precision in app.js?

klassenjs:
To answer my own question: Yes.

PR #637, opened 2021-03-16
Update tests to use Ubuntu 20.04

klassenjs:


klassenjs:
`ubuntu-latest` seems broken at the moment and will be transitioning to 20.04 soon anyway.  Probably better to explicitly specify the version anyway.

Issue #365, opened 2019-03-01
Buffering a selected polygon doesn't seem to work


tchaddad:
As mentioned in https://github.com/geomoose/gm3/issues/48, selecting a polygon, then applying a buffer, in order to select neighboring properties, is a common workflow.

Attempting to do this in the demo, here is what I do:

- Step 1: Select polygon with point
- Step 2: Return to select menu, and switch to "using Feature from Results", and "with buffer 300 feet"
- Step 3: press Go

The result is that the the buffer is applied to my point, and not to the selected polygon, and so the number of selected parcels is incorrect. In fact, as shown in the screenshots below, for a parcel with dimensions larger than 300 feet, it is possible that the buffer lies entirely inside the parcel, and so no neighbors are selected.

Is my method incorrect, or is buffer working on the wrong object?

_____________

**Step 1**
![image](https://user-images.githubusercontent.com/1042618/53617803-a2475980-3b9d-11e9-8b41-155f36b320d8.png)

_____________

**Step 2**
![image](https://user-images.githubusercontent.com/1042618/53617828-be4afb00-3b9d-11e9-8477-56976dbe3910.png)

_____________

**Step 3**
![image](https://user-images.githubusercontent.com/1042618/53617838-d6227f00-3b9d-11e9-8816-21032e42381c.png)

klassenjs:
Looks like you didn't click on the parcel to select it (after choosing select from results).  The parcel shouldn't be yellow.

tchaddad:
Ok - thanks, that's helpful, although not necessarily intuitive. How does the user know that they need to select something again that looks like it is already selected?

Not arguing, just making a case for a better narrative so that users don't make this mistake.

Where would I look to edit the instructions that show up in the select panel? I'll see if I can come up with language that can be put into a pull request...

klassenjs:
One thing that might be a bug is it "remembering" what was previously selected (the point from the previous select).  If it didn't do that, it would give you an error telling you that you need to select something.

Here is where to start looking: https://github.com/geomoose/gm3/blob/master/src/services/select.js

klassenjs:
Although that remembering can also be really useful if you want to re-run your last query with different options.

chughes-lincoln:
Is there any way currently to buffer multiple results? This appears to only let you do a search or a selection, and then buffer one of the results. It would be useful to be able to buffer multiple parcels (ie in a certain subdivision or zoning) rather than one parcel.

theduckylittle:
[...]

PR #643, opened 2021-05-09
Add "Buffer All" functionality

theduckylittle:
Adds a new "Buffer All" tool to the Super Tab results
when configured.

Bonus! Fixes bug with buffer-select not showing grid
results.

### DIFF
--- a/docs/howto/customize-results.rst
+++ b/docs/howto/customize-results.rst
@@ -1,0 +1,35 @@
+How-to customize result tools
+#############################
+
+GeoMoose can be configured to off the user additional
+options for results in the Super Tab.
+
+Service results configuration options
+-------------------------------------
+
+When adding a service, administrators can configure which tools
+are available to a user using the ``results`` configuration object.
+
+Here is an example adding buffer all and removing the count
+of layers:
+
+::
+
+    app.registerService('select', SelectService, {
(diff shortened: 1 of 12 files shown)


chughes-lincoln:
I've tested this out on my build and everything is working great. This, combined with the ability to buffer from a single result, may for the most part address https://github.com/geomoose/gm3/issues/365. The default buffer tool can do some more advanced things like buffering a sketch, or buffering a subset of your results, which still isn't very intuitive to the average user, but I think there are now relatively straightforward ways for admins to configure simple buffer applications.

klassenjs:
It appears this broke Search Parcels -> Owner: Smith -> Go causes a JS error

chughes-lincoln:
Not sure how I missed the search not working. I'm also having issues with larger buffers. On my build of the demo,
I selected around 300 parcels and tried to create a 10 foot buffer, which returned no results.

theduckylittle:
Fixed the search error. The search service itself also had some crusty code which has been updated.

@chughes-lincoln - We look to be hitting a limit in MapServer:
```
msOWSPreParseRequest(): OWS error. XML parsing error: Excessive depth in document: 256 use XML_PARSE_HUGE option
````

PR #644, opened 2021-05-10
Add Bounding Box and Z/Lon/Lat support to loc param

theduckylittle:
Allow passing in more location formats:

A. `zxy` - the default GeoMoose format since 3.0.
    This is really `map resoltion;center x;center y`.
B. `lonlat` - Which is formatted as `zoom/lon/lat`.
C. `bbox` - Formatted as `west;south;east;north`. In ESPG:4326.

Adds the ability to change the hash emit format as well using
the new option for the HashTracker class. In `app.js`:

```
    var hash_tracker = new gm3.trackers.HashTracker(app.store, {
        locationFormat: 'lonlat'
    });
```

### DIFF
--- a/src/gm3/actionTypes.js
+++ b/src/gm3/actionTypes.js
@@ -94,6 +94,7 @@ export const MAP = {
     REMOVE_FILTER: 'MAP_REMOVE_FILTER',
 
     SET_EDIT_PATH: 'MAP_SET_EDIT_PATH',
+    RESIZE: 'MAP_RESIZE',
 };
 
 export const TOOLBAR = {
--- a/src/gm3/actions/map.js
+++ b/src/gm3/actions/map.js
@@ -202,3 +202,17 @@ export function setEditPath(editPath) {
         editPath,
     };
 }
+
+/* Set the size of the map in the state as a hint
+ * to other components.
+ *
+ * @param size {Object} an object containing a width and height property.
+ *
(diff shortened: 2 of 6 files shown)


chughes-lincoln:
Is it intended that the locationFormat needs to be declared in the app.js for the default xyz format to work? When I tested this on my build, if I have a service declared to start on the initial load (via `app.startService('simple-search');`) I get an uncaught type error and the service doesn't automatically start (It still works if I manually run the tool, it just doesn't start on startup). Everything else appears to work as normal. If I then specify the locationFormat in the app.js (whether it's xyz, bbox, or latlon), everything works as it should.

theduckylittle:
I'll check on the error.  Not sure how to note this appropriately but:
1. ON STARTUP: The application will accept any of the three location formats (zxy, lonlat, bbox). It will check to see which one is coming in and use that as the startup-location.
2. ON NAV: The application will put in the hash whichever is configured in `locationFormat`. Without being specified, `zxy` should work. I would be willing to hear-out an argument that we should switch to `lonlat` as the default format as I think that is more accessible, sensible, shorter, and more common.

theduckylittle:
Fixed the startup parameter issue! Needed to ensure state was ready before formatting!

chughes-lincoln:
Everything looks good now, thanks!

Issue #640, opened 2021-04-14
Print Inconsistency - As you turn on more layers, some don't appear in the print out


chughes-lincoln:
Using the demo as a test case, when I print with just the parcels layer I don't notice any issues.
When I turn on the AGS County Rail and County Runways, some layers fail to show up in the print out.
In this case the parcels appeared for a second in the print preview, and then disappeared.

![image](https://user-images.githubusercontent.com/43766648/114760425-245e5200-9d14-11eb-93e3-8725218f15a8.png)

brentfraser:
Works fine for me (Firefox v87.0 64-bit) using the full demo on geomoose.org.  Anything special on your client -side?

chughes-lincoln:
Interestingly, I've now only managed to replicate it from my work computer (Windows). The only extension I can see is is Virtru Email Protection, which I wouldn't think would have an impact.
-Chrome - Version 89.0.4389.114 (Official Build) (64-bit)
-Microsoft Edge - Version 89.0.774.68 (Official build) (64-bit)

For the most part it works fine from my Mac. I was able to replicate it by turning on the OpenStreetMap layer (the OpenStreetMap shows in the print, but the parcels do not), but I was able to turn on the ags railroads/runways (in addition to a bunch of other layers) without issues.
-Chrome - Version 89.0.4389.114 (Official Build) (x86_64)
-Safari - Version 14.0.3 (16610.4.3.1.7)
-Firefox - Version 86.0.1 (64 bit)

chughes-lincoln:
Some more diagnostics:

My office desktop computer (Windows, same browser settings and extensions as my work laptop) works perfectly. I can turn on all the ags layers and print them without issue, I can even turn on the OpenStreetMap layer and print it, which wasn't working on my Mac.

chughes-lincoln:
I also have similar issues from my iPhone.
I can turn on the two ags layers I mentioned without issues, but if I turn on OpenStreetMap, only
the OpenStreetMap layer shows in the print preview.

brentfraser:
I was trying it with Chrome and I see there could be a Z-order / Layer ordering problem.  The ArcGIS background layer obscures the AGS Rail and Runway layers

![image](https://user-images.githubusercontent.com/1638313/114789780-dca1ef80-9d40-11eb-8f0b-dbf8d77d566d.png)


May be related.  Or not

chughes-lincoln:
[...]

PR #649, opened 2021-05-24
Fix for high resolution print

theduckylittle:
1. Fixes the canvas selector based on the OpenLayers demo code
   and tweaks from robinrullo.
2. Refactors and cleans up components for print.
3. Adds a progress bar which is handy for when big images
    are coming in over the pipe.

refs: #640

### DIFF
--- a/src/gm3/components/linearProgress.js
+++ b/src/gm3/components/linearProgress.js
@@ -1,0 +1,10 @@
+import React from 'react';
+
+const LinearProgress = () => (
+    <div className="linear-progress">
+        <div className="bar">
+        </div>
+    </div>
+);
+
+export default LinearProgress;
--- a/src/gm3/components/map/index.js
+++ b/src/gm3/components/map/index.js
@@ -967,7 +967,7 @@ class Map extends React.Component {
 
         // call back for when the map has finished rendering.
         if(this.props.mapRenderedCallback) {
-            this.map.on('postrender', this.props.mapRenderedCallback);
+            this.map.on('postrender', () => this.props.mapRenderedCallback(this.map));
         }
 
         // once the map is created, kick off the initial startup.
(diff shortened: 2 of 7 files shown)


robinrullo:
I think that only printing the first layout works: the exportSize of getImage seems to not be updated when another layout is selected. I tried printing page-letter-portrait, the map size is good but the canvas size is not. 

To reproduce, select page-letter-landscape (without legend because the map size format of page-letter-landscape-legend is similar as portrait), close modal, reopen print (that will set the canvas size to page-letter-landscape-legend format) and then select page-letter-portrait, you will see that: 
image
image

theduckylittle:
Found it! There was an issue with the map-rendering callback creating a closure.

robinrullo:
There is another issue!
To reproduce, just change the size of the browser window (or open console does the trick) when you are in printmodal.
image
image

theduckylittle:
Thanks for the great testing work on this!

I wasn't able to recreate the state error bug but I was able to recreate an infinite loop. A "test fix" I did last night was causing an issue. The last commit should fix it. If you're having problems on your latest build, let me know.

robinrullo:
Everything seems to be working fine now 😃
There is just one thing (I don't know if it's done on purpose), the printpreview no longer adapts to the canvas size.

Before: 
image

Now:
image

Glad I can help improve the app  !

theduckylittle:
And now that is fixed too.

robinrullo:
I accidentally opened/closed very quickly the modal when the map was still being rendering in the preview and got the same state error than before

theduckylittle:
Oy! Okay. I think I got it for real, finally, completely this time. I wrapped the `setImage` call in a check to make sure the `parentRef` still existed.

chughes-lincoln:
I rebuilt the demo on this PR and so far I'm not seeing any of the errors that were mentioned. It successfully fixes the issue of certain layers (sketches, some basemaps) not appearing in the print preview or the print out when using certain high resolution monitors. 

The only odd thing is that I haven't been able to get that progress bar that you mentioned to show.

Issue #632, opened 2021-02-24
Select - Enter Keyboard Shortcut


chughes-lincoln:
I'm not sure if this previously worked (See https://github.com/geomoose/gm3/pull/119 and https://github.com/geomoose/gm3/issues/61), but pressing Enter for select services doesn't work as a shortcut for 'GO'.
It does work for search services.

chughes-lincoln:
Enter for select works with this PR. 

What could cause issues for users, is that pressing Enter when a tool isn't the active tab still runs that tool (i.e. if you have a service set to run on startup and switch to the Catalog, or if you clicked the select tool and then switched to the Catalog, hitting Enter runs the tool that was on in the background). 

For searches this isn't much of a problem because the program catches when someone tries to run a search w/o an input; for selections, a selection is run with an empty geometry which selects the entire dataset, potentially causing the page to crash if the dataset is large enough (You could already do this by clicking 'GO' before inputting a geometry, but it's a bit easier to unintentionally press Enter, than unintentionally click 'GO', particularly if you're not on the Super Tab).

PR #659, opened 2021-06-06
Restore keyboard short cuts for service form

theduckylittle:
1. Put the keyboard shortcuts on the document instead
   of on the form to make Escape and Enter work properly.
2. Fix state-up issue with `submit()` message.

refs: #632

### DIFF
--- a/src/gm3/components/serviceForm.js
+++ b/src/gm3/components/serviceForm.js
@@ -1,7 +1,7 @@
 /*
  * The MIT License (MIT)
  *
- * Copyright (c) 2016-2017 Dan "Ducky" Little
+ * Copyright (c) 2016-2017, 2021 Dan "Ducky" Little
  *
  * Permission is hereby granted, free of charge, to any person obtaining a copy
  * of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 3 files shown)


chughes-lincoln:
Enter for select works with this PR.

What could cause issues for users, is that pressing Enter when a tool isn't the active tab still runs that tool (i.e. if you have a service set to run on startup and switch to the Catalog, or if you clicked the select tool and then switched to the Catalog, hitting Enter runs the tool that was on in the background).

For searches this isn't much of a problem because the program catches when someone tries to run a search w/o an input; for selections, a selection is run with an empty geometry which selects the entire dataset, potentially causing the page to crash if the dataset is large enough (You could already do this by clicking 'GO' before inputting a geometry, but it's a bit easier to unintentionally press Enter, than unintentionally click 'GO', particularly if you're not on the Super Tab).

theduckylittle:
I found that the service's check for an empty-geometry was broken! That would've been a bug that changed in 3.7 when the way an empty-selection was refactored. I updated the select.js service code to appropriately test for whether there is a selection or not.

chughes-lincoln:
Works on my build now! That alleviates my concerns. It's probably rare for someone to press enter with a select service in the background, but it could still have had some pretty bad consequences if it was selecting entire datasets.

Issue #650, opened 2021-05-25
Sketch drawing issue: The point disappears when a label is added to it


robinrullo:
From "Drawing and Markup", when a label is added to a point, the point disappears due to `declutter` (Set to `true` for all vector layers).
I tried to put a text-anchor from mapbox style but it doesn't seem to shift the label enough to make it reappear.

This should be a point with a label:
![image](https://user-images.githubusercontent.com/25304220/119510701-09025000-bd72-11eb-80e4-ecbc01b74372.png)

Edit: This is a known issue of ol-mapbox-style: https://github.com/openlayers/ol-mapbox-style/issues/279

PR #660, opened 2021-06-06
Remove forced declutter and move it to source config.

theduckylittle:
To restore the declutter behaviour,

```
<map-source ...>
    <config name="declutter">true</config>
    ...
</map-source>
```

refs: #650

### DIFF
--- a/docs/howto/add-decluttering.rst
+++ b/docs/howto/add-decluttering.rst
@@ -1,0 +1,13 @@
+How-to add decluttering to a vector layer
+=========================================
+
+Decluttering is offered by OpenLayers and will create less crowded
+maps by not rendering features which overlap. As of GeoMoose 3.8.0,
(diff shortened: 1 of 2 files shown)


chughes-lincoln:
Works as expected on my build, though I wasn't able to reenable clutter in the map-source until I set it to,

```
<map-source ...>
    <config name="declutter" value = "true"/>
    ...
</map-source>
```

theduckylittle:
Ack. My bad. Got too punchy with my descriptions. Added a how-to which _correctly_ details how to restore decluttering.

klassenjs:
This one is pushing it for not being a breaking backwards compatibility by changing the default behavior.  But, maybe it is a bugfix + new feature.  Whatever.

theduckylittle:
Underlying library changed behaviour this is a catch-up/bug-fix plus new feature.

Issue #658, opened 2021-06-02
Sketch Editing - Attributes not retained on initial sketch


chughes-lincoln:
Testing on the online demo, when adding a sketch, the default options show up in the Edit Feature Properties modal that pops up (color, fill, opacity, etc...). However, when clicking Edit feature properties, none of the attributes of that sketch are retained (If you use the defaults when making the initial sketch - any changes you make in the initial pop-up are retained), making it a bit harder to make adjustments (i.e. if you wish to make a polygon slightly more opaque, you don't necessarily know what the current opacity is, and you need to re-specify the fill color to change the fill opacity). Though if you do make changes these attributes are retained if you return to edit the same sketch. 

None of this prevents you from doing anything as you can still make any changes you'd like, but it makes it a bit less user friendly.

PR #661, opened 2021-06-06
Ensure default properties are saved to property

theduckylittle:
While property changes were saved, when recalled they
would be undefined, making it difficult to determine
their rendered values. This change requires that all
defaults be set by the properties.

refs: #658

### DIFF
--- a/src/gm3/components/editor/modal.js
+++ b/src/gm3/components/editor/modal.js
@@ -5,8 +5,16 @@ import Modal from '../modal';
 
 const isNumberType = type => (type === 'number' || type === 'range');
 
-const getDefaultValue = attr =>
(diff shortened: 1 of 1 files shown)


chughes-lincoln:
Testing on the demo with this PR, it looks like it's no longer recognizing some of the default values, specifically stroke size, stroke opacity, and fill opacity. If you adjust those values in the initial sketch creation (even if you adjust them and then return them to the defaults before you click 'Save Changes'), the opacity/size values are retained, but not if you just accept the defaults.

theduckylittle:
Fixed! There was an issue with getting the default value if it was a numeric type.

chughes-lincoln:
Everything looks good now!

Issue #663, opened 2021-06-08
Special characters not correctly sorted in the Grid


robinrullo:
Values in the Grid are not sorted correctly when words begin with special characters. Currently, special characters are sorted after **Z** (in javascript, compare `'É' > 'Z' = 1`). But **É** should be between **D** and **F**.
And also in Swedish, **Ä** is after **Z** but in French is before **B**.
So [`Intl.Collator().compare(a, b)`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Collator/compare) can be used in this case and use the browser locales.
```diff
sortResults(results) {
    const sorted_results = [].concat(results);

    const sort_col = this.state.sortBy;
    const sort_asc = this.state.sortAsc;
    const sort_as = this.state.sortAs;

+   const collator = Intl.Collator();
    sorted_results.sort(function(a, b) {
        let value_a = a.properties[sort_col];
        let value_b = b.properties[sort_col];
        if(sort_as === 'number') {
            value_a = parseFloat(value_a);
            value_b = parseFloat(value_b);
        }
        if(sort_asc) {
-           return (value_a > value_b) ? 1 : -1;
+           return collator.compare(value_a, value_b);
        }
-       return (value_a > value_b) ? -1 : 1;
+       return collator.compare(value_b, value_a);
    });
    return sorted_results;
}
```

elil:
See also #378

Issue #378, opened 2019-04-29
Sort grid results occasionally fails


chughes-lincoln:
Sorting grid results on the 3.3.1 demo occasionally fails. It appears to be isolated to fields that have null values. I.e. if you select  all the parcels and then sort by City, it's out of order. 

It works without any issue on my fields that aren't missing any data, but not the ones that return ??? values.

PR #665, opened 2021-06-15
Fix for non-english and null values in grid sort

theduckylittle:
Using `Intl` fixes both #663 and #378.


klassenjs:
This breaks sort by Est Value for parcels.  It appears to be sorting alphabetically instead of numerically.

Issue #564, opened 2020-06-30
Set Initial Extent by Bounding Box


chughes-lincoln:
Currently (unless I'm understanding things poorly), the initial extent of the application is set in your app.js using a centroid and a zoom parameter.

```
app.loadMapbook({url: 'mapbook.xml'}).then(function() {
    // set the default view.
    app.setView({
        center: [ -10370351.141856, 5550949.728470501 ],
        zoom: 12
    });
```

This works for the most part, except it forces you to make an educated guess regarding what the screen size of your users are (or even if they have their window maximized - e.g. using the demo on my laptop results in a map that is a bit cropped when looking at the parcel data)

It would be useful to be able to set an extent parameter in the app.js for the initial extent, similar to when using the zoom to full extent tool:

```
    app.registerAction('fullextent', ZoomToAction, {
        extent: [-10742765,5398288,-9920914,6310641]
    });
```

chughes-lincoln:
Fixed by https://github.com/geomoose/gm3/pull/676

PR #676, opened 2021-08-01
Fix state bug with zoom to extent and document

theduckylittle:
1. Fixed a startup-view bug with `zoomToExtent`.
2. Added documentation on how to start from an extent.

refs: #564


chughes-lincoln:
All of this works as expected on my build, thanks!

Issue #636, opened 2021-03-16
Discussion: Re-organization of Layer tools


brentfraser:
(This issue is about tools on the layers in the Catalog tab, not the page toolbar or tools on the map.)

**The Problem:**

  The layer tools add visual clutter to the Catalog's layer list.  While the tools add access to functionality, their appearance on every layer (or at least those configured by the admin) can cause cause confusion/anxiety by the user.

**The Solution:**

In keeping with the concepts of information hiding and minimizing chartjunk, I propose we hide the tools in two toolbars: one for "standard" tools, and one for editing tools.

_The current situation:_
Note: because all the tools are enabled in the mapbook (not a common situation), there are lots of them shown, so they "wrap" to fit in the Catalog tab.

![image](https://user-images.githubusercontent.com/1638313/111329285-a0da1400-8634-11eb-8b07-a4a7908f79dd.png)

_The proposed solution:_
In the solution shown below, the Standard toolbar is shown/hidden by clicking on the Layer title, while the Edit toolbar is shown/hidden by clicking on the "edit" icon to the left of the Layer title (if editing is configured in the mapbook).

![image](https://user-images.githubusercontent.com/1638313/111352268-9bd38f80-8649-11eb-84ba-b1d5c697b5b2.png)

Comments appreciated!

elil:
Generally seems to make sense.

The edit icon is very obvious despite the fact that it is the less common
situation.  Would it make sense to swap the function of clicking the layer
for edit and a (different) icon for show tools?  Or keep the edit icon and
use a different indicator for the availability of more tools (bold layer
title, layer title with grey highlight behind it, something else)?

Eli

brentfraser:
@elil Yes displaying the edit icon would be rare, since most installations would not have editing layers.  And we could have an icon to indicate tools-are-available-for-this-layer.  I had that in my original proposal but instead went for a minimalist look.

brentfraser:
At our last PSC meeting Jim had asked about putting the edit icons on the map like the zoom in/out icons (and the edit geometry icons).  Here's a mock up of what that could look like:

![image](https://user-images.githubusercontent.com/1638313/126196456-0581c825-f3d0-4787-8743-0afd8f30cbd4.png)

Note:
- the edit icons are removed from the catalog
- the enable-editing icon is now on the left of the layer title in the catalog

elil:
[...]

PR #677, opened 2021-08-02
UX Updates for Editing Tools

theduckylittle:
1. Clean up rendering logic for drawing tools.
2. Move the drawing tools to the map from the catalog.
3. Create a "edit trigger" from the catalog.
4. Allow adding classes to catalog layers and add an inline tools style.
5. Additional clean up of catalog layers
6. Better wording for opening the drawing tools + translation

### A single "edit" tool

image

### Opens a toolbar on the map

image

### With a tool selected, show the stop sign

This collapses down to just the stop sign. Then returns back to the toolbar when "stop" has been clicked.

image

refs: #636


brentfraser:
Is there more to come?  The Stop button doesn't actually save the edits, etc.

theduckylittle:
It never did?

This hasn’t changed the behavior of any of the tools.

brentfraser:
How do I save my edits?

brentfraser:
Are we missing the "Save" and "Undo" buttons?

robinrullo:
You mixed up the keys used for paths but I think to draw, you use `activeSource` and to then edit the feature, you use `editPath`.
There is used `editPath`: https://github.com/geomoose/gm3/blob/bd06cf76f36ee6575e718fd523f404ee3f69d0ba/src/gm3/components/catalog/tools/edit-tool.js#L48
and then is `activeSource` instead of `editPath`:
https://github.com/geomoose/gm3/blob/bd06cf76f36ee6575e718fd523f404ee3f69d0ba/src/gm3/components/map/index.js#L1328

So Drawing from `Drawing and Markup`, is drawing to selection layer (activeSource = '' => path = null => selection layer).

theduckylittle:
Oopsy doodle. 

I see where I messed up, I had forgotten to pass in the active source when changing the draw tools. Things look to be working more properly now!

brentfraser:
Much better.  Is there someway we can enforce "the-layer-must-be-on-to-edit".  Otherwise strange things happen.

brentfraser:
Or just exit the edit tools when the layer is turned off.  (Warn if there are pending edits?)

theduckylittle:
Done and done. The user can still hurt themselves by turning off the layer while editing. I'm willing to ticket that for follow up work if folks really think it'll be an issue.

klassenjs:
This has some strange interactions with the select tool.

Edit Drawing and Markup.  Add a Polygon (doesn't really matter what)
Select (note don't click stop)
Can draw stuff normally for select
Click Stop
Now the Select tool and the Drawing and Markup tool (on map) are linked (changing tools on the select tab or map seem to change the other)
Now draw something (seems to go to the Drawing and Markup layer if picked the tool from the map and goes to the select tool geom if picked the tool from the tab.)

brentfraser:
@theduckylittle  Re edit tool in catalog.  Can we move the edit tool to be to the left of the favourite tool (so between the fav tool and the layer name).  That way we could get rid of the white space caused by having the edit tool on a separate line/div.

klassenjs:
Continued in #681

Issue #673, opened 2021-07-29
Print: bad scalebar width according to the unit 


robinrullo:
Using page formats having different units than inches, the width of the scalebar is completely wrong.

It's calculated as if unit is still inches:
https://github.com/geomoose/gm3/blob/29e2626dc240ff664a74ed6f00f1a3e36434b5b5/src/gm3/components/print/printModal.js#L251

instead of using the px to layout coef:
```js
width: scaleInfo.width * pxToLayout,
```

theduckylittle:
Silly Ami's and our forgetting to test in metric...

PR #678, opened 2021-08-03
Use pxToLayout units instead of just inches

theduckylittle:
HT to @robinrullo for the fix.

refs: #673

### DIFF
--- a/src/gm3/components/print/printLayouts.js
+++ b/src/gm3/components/print/printLayouts.js
@@ -154,6 +154,50 @@ const LAYOUTS = [
                 x: .5, y: 10.5, text: 'Printed on {{month}} / {{day}} / {{year}}'
             }
         ]
+    },
+    {
+        label: 'a4-landscape-legend',
+        orientation: 'landscape',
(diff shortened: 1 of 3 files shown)


klassenjs:
I don't think this is right.  2mi scalebar vs 1mi grid.  Scalebar looks like it is closer to 2.5mi.

image

Old version:
image

klassenjs:
I'm getting pxToLayout = 1/54 when I suspect it was intended to be 1/72 for the unchanged case.

klassenjs:
Also, can we add a metric paper size (A4?) to the demo to make testing stuff like this easier?

klassenjs:
It looks to me like in `toPoints` the `case 'px'` should also be 72.  I'm getting scale bars that are 4/3 the length they should be which happens to also be 96/72.  Curious the right value is system dependent.

klassenjs:
Either that or the 'px' is correct, but the pxToLayout in the scalebar code should really be ptToLayout.  I'm really not sure which is correct.

PR #687, opened 2021-09-15
Add some tool tips to the English and Spanish json

chughes-lincoln:
I would have added these to the other language files, but I'm not confident enough to make anything other than google translations.


theduckylittle:
Merged here: https://github.com/geomoose/gm3/pull/693

Thank you!

PR #690, opened 2021-11-08
Update tests to build with nodejs 16.x

klassenjs:
Prepare for NodeJS 16.x

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -37,7 +37,7 @@ jobs:
       fail-fast: true
       matrix:
         platform: [ 'ubuntu-20.04' ]
-        node-version: [ '14.x' ]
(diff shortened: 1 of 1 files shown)


theduckylittle:
What about switching to the latest LTS to keep us honest? 

https://github.com/actions/setup-node#supported-version-syntax

That way any MR will get flagged if we have issues with the latest LTS of node.

klassenjs:
I mean that is basically what I am doing by hand.  I think at one point we said we'd support the current nodejs LTS and the previous LTS.  In practice we are only testing and building against one version.

The one benefit I see of specifically pinning the node version is that if/when we go looking though history, it is explicit which node version we were using at the time.  I'd be willing to go either way though.

klassenjs:
Replaced by #691

Issue #697, opened 2022-01-06
Filter when selection set has any null values


chughes-lincoln:
Attempting a filter using the grid/table when the selection set has any null values in that field causes the map to crash.

To replicate in the online demo, run a search so that some features have null values in the City field, then attempt to filter the City field.
It won't let you uncheck the features with null values. If you uncheck features w/o null values and then click 'Set', you get a blank map and a console error stating there was an invalid value.

Editing the source data to N/A or something else fixes the issue, though that's not always feasible or desired.

chughes-lincoln:
Just a summary of the current state:
-you can now filter out null values in a selection set
-the map still crashes if you attempt to filter out non-null values (if the filter field contains null values)

PR #701, opened 2022-01-11
Fix issue with undefined values in the grid

theduckylittle:
`undefined` was not a supported value by the mapbox filter
generator. This updates the query to force undefined to become
`''` (an empty string).

refs: #697

### DIFF
--- a/src/gm3/components/grid.js
+++ b/src/gm3/components/grid.js
@@ -63,7 +63,21 @@ class FilterModal extends ModalDialog {
         const new_filters = [];
 
         if(this.props.column.filter.type === 'list') {
-            new_filters.push(['in', property].concat(value));
+            // undefined causes challenges for the filter generator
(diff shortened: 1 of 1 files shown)


chughes-lincoln:
I'm still getting this error when I rebuilt the demo on this PR.
This time the null values show as (empty). Trying to uncheck the null features fails
fails, and hitting 'Set' causes the map to crash.

theduckylittle:
Try this update! Took an alternative approach to generating the list options.

klassenjs:
@chughes-lincoln Are your interesting errors caused by this PR or is it a separate bug?

chughes-lincoln:
It's probably a separate, related bug. Though it only seems to manifest itself when there's a null value in the selection set.
Any case where I can replicate it on this PR would also fail on the demo because there's a null value.

PR #707, opened 2022-02-16
Update tooltips

chughes-lincoln:
The Unfade tool references unfade-layer, not unfade-tip


theduckylittle:
Ack! Corrected the code to use `unfade-tip` instead of `unfade-layer`. 

Two reasons for doing that:
1. Makes for a smaller diff.
2. The rest should have "-tip" in their translation name which better indicates where they should be.

chughes-lincoln:
Thanks! As I was submitting the PR I realized yours was the better route but I was already commited.

Issue #706, opened 2022-02-16
Legend settings from Catalog not retained in Print


chughes-lincoln:
You can turn off legends in the Catalog by setting show-legend="false" in the mapbook.xml. Turning off legends in the Catalog does not also turn them off in the print legend. Removing the class name in the mapfile does remove the legend in the print, but it leaves an empty gap where the legend would have been.

There's a similar issue with scale-based layers. If you're at a scale greater than the max scale of the layer (with legends enabled), an empty gap shows in the legend (rather than no legend at all due to the layer not being visible at that scale). This empty gap appears in the legend.

This isn't really an issue for the demo, but for sites with more data and more complex layers that have multiple classes or have scale based settings, it results in a lot of empty gaps in the print legend, which makes it much more difficult to fit the entire legend in the print.

chughes-lincoln:
Example of what I'm referring to when the legend is off in the catalog and there's no CLASS Name in the mapfile (ignore the poorly formatted legend. I was adjusting sizes and saw that the empty gaps I was referring to show up against a background).

<img src="https://user-images.githubusercontent.com/43766648/154343457-7b140d9b-5ad7-4026-aed7-16f08f9babe1.png" width="600" height="800">

PR #708, opened 2022-02-17
Fix print legend visibility issues

theduckylittle:
1. Adds check to ensure min/maxresolution are in range when
   specified.
2. Adds a check to ensure the legend has not been turned off in the
   catalog.

refs: #706

### DIFF
--- a/examples/desktop/mapbook.xml
+++ b/examples/desktop/mapbook.xml
@@ -171,13 +171,11 @@
     </map-source>
 
     <!-- Demo of parcels as points. -->
-    <!--
-    <map-source name="parcel_points" type="mapserver" up="true" down="true" title="Parcel Points">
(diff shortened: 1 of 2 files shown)


chughes-lincoln:
This makes my print legends look so much better already!

A couple of notes:
1) It correctly honors legends being set to false in the catalog
2) It doesn't seem to honor scale based legends in the catalog or the print out (not a new issue in the catalog, I believe that's how it already was)

An example for 2: I have a contours dataset with 40ft and 10ft contours grouped into the same map-source. At the full extent and scale of my map neither of them show up in the map, but they do show up as white space in the catalog legend and the print legend. If I zoom in a bit the 40ft contours appear, including the legend. At this scale there's some white space in the catalog and print legends due to the 10ft contours. If I zoom in further the 10ft contours replace the 40ft contours, and I now see some white space in the catalog and print legends due to the 40ft contours.

PR #713, opened 2022-03-10
Add ability to create a search example

chughes-lincoln:
This allows admins to configure examples for the searches so users need to spend less effort figuring out the format of the data:

image

### DIFF
--- a/examples/desktop/app.js
+++ b/examples/desktop/app.js
@@ -101,7 +101,7 @@ app.loadMapbook().then(function() {
 
     app.registerService('search', SearchService, {
         fields: [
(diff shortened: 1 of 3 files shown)


theduckylittle:
I like this but it might be helpful for us to follow the example laid down by material ui:
-  https://mui.com/components/text-fields/

Any objections to calling the property `helper` and moving the text below the input field?

chughes-lincoln:
That's completely fine by me.

klassenjs:
There is also the HTML placeholder attribute:

```
<input type="text" placeholder="example text">
```

theduckylittle:
Closing in favor of work done on #722

Issue #711, opened 2022-02-22
Large Buffer returns no results; Really Large Buffer crashes map


chughes-lincoln:
The former is a known issue, but it's related so I'm adding it here. I'm not sure what could be done to address it, perhaps doing a union of the selection set before the final buffer, or warnings that large buffers aren't supported so the user knows what the issue is.

If you do a moderately large selection (i.e. the 300 parcels in the middle of the demo), the buffer returns no results and you get a mapserver error: msOWSPreParseRequest(): OWS error. `XML parsing error: Excessive depth in document: 256 use XML_PARSE_HUGE option`. 

If you do a really large buffer (i.e. select almost all of the parcels in the demo), you a larger error along with a map crash, forcing you to refresh your browser to do anything: `Uncaught RangeError: Maximum call stack size exceeded`

chughes-lincoln:
If you overwrite the bufferMaxFeatures value in https://github.com/geomoose/gm3/pull/715 it looks like large buffers are now successful due to unioning the buffer features before running the buffer.

Though creating the union on the fly is pretty data intensive so you wouldn't necessarily want to run a large buffer (i.e. I can run a buffer around 1000 features, but when I set the buffer size of 1ft, it hangs for a while while it's creating the union).
My only thought for a potential enhancement would be waiting to create the union of features until the user runs the buffer,
so that it's not changing every time the user changes the buffer size.

PR #715, opened 2022-03-18
Limit the number of bufferable features

theduckylittle:
The browser can only handle so much, limit the number of
features GeoMoose willingly buffers from the results.

refs: #711


klassenjs:
Is there a clean way we could tell the user why the buffer button is disabled?

theduckylittle:
Could put a message on the bottom. Would need to translate that message to our various languages :grimacing:

klassenjs:
Think it is worth it or just merge it?  As is it could just look broke without saying what could fix it.  Although that is still better than crashing.

elil:
I think that things like this do well being very interactive.  Allow people
to hit the button but the first thing is a check with a very obvious
message, "Buffer limit of # exceeded"

theduckylittle:
Yeah... I'm sympathetic to giving the user too much information. I've updated this to show a modal when there are too many features.

PR #716, opened 2022-03-18
2022 March Audit

theduckylittle:
Updates a collection of the dependencies.

Stay healthy.


klassenjs:
The new build has [0-4].geomoose.js and [0-4].geomoose.min.js files.  I'm not sure if they do anything or not.  geomoose.js and geomoose.min.js are still there.

theduckylittle:
Previous comments were fallout from the Webpack5 update. I _think_ everything should be fine when released to production but we might need to wait until we hear back from users re: updating an application. 

This update is clean but ESLint will throw a bunch of (actually quite useful) warnings.

I want to address line warnings in a separate PR.

PR #718, opened 2022-03-18
Switch to Webpack 5

theduckylittle:
- Updates to new webpack version
- Cleans up old deps on the configuration for the dev server.


klassenjs:
First impression is it seems to work and the app feels faster.

I had an issue with the mobile app, but it was a local issue. I hit one weird state thing after semi-randomly exercising the desktop example for awhile, but couldn't reproduce it.  It's Probably not related to this PR.

PR #720, opened 2022-03-22
Add documentation for bufferMaxFeatures

theduckylittle:
- Adds docs for bufferMaxFeatures option.
- Updates modal to render the number of max features.

### DIFF
--- a/docs/howto/configure-max-bufffer.rst
+++ b/docs/howto/configure-max-bufffer.rst
@@ -1,0 +1,28 @@
+Configure Number of Buffer-able Features
+========================================
+
+Update the `app` declaration in `app.js`. Include the `query` configuration object,
+and set `bufferMaxFeatures` to the largest number of features the user should
+be allowed to buffer.
+
+::
+
+    var app = new gm3.Application({
+        mapserver_url: CONFIG.mapserver_url,
+        mapfile_root: CONFIG.mapfile_root,
+        map: {
+            scaleLine: {
+                enabled: true,
+                units: 'imperial'
+            }
+        },
+        mapbooks: {
+            'default': 'mapbook.xml',
+            'editing': 'mapbook-editing.xml',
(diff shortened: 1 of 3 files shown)


chughes-lincoln:
The issue with larger buffers is that a union of the features is being created when the user inputs the buffer size. When I do a buffer of around 400 features I start getting alerts that my browser is unresponsive (though it eventually finishes and then I can run the buffer). So I'd assume buffers much larger than that wouldn't be viable. On a whim I tried a buffer of 40,000 features, updating the With Buffer parameter to 5ft ran for 15 minutes before I got impatient and gave up.

brentfraser:
What are your client computer specs?  o/s? ram?  cpu?

chughes-lincoln:
Windows 6-core@3GHz, 32GB RAM

brentfraser:
That is likely better than most users.  I use a 3 second rule/guideline for GUI based response.  If there's not feedback within 3 seconds, users tend to lose their workflow momentum.   What value for bufferMaxFeatures  would result in a 3 second buffer time on your client machine?

chughes-lincoln:
I'd say around 200 features.

klassenjs:
One thing to keep in mind is number of features is what we can easily
check, but number of vertices is what really drives the computational
complexity.

One feature with a ton of vertices like one following a river could be
worse than 100s of 4 point polygons.

So given variations in datasets and in client compute power and memory
availability, the “right” max features number to achieve a worst case
performance target is going to vary case to case.

The right thing to do might be to put the calculation in a service worker
(to get it out of the UI thread) and set a timeout to cancel the task if it
takes over 3 seconds to complete.  I’m not sure what it would take to
implement that, but I expect it wouldn’t be easy.

klassenjs:
To be clear, my position on this PR is that it makes GeoMoose better.  Just because we could potentially do something even better in the future shouldn't block merging this now.

PR #735, opened 2022-06-29
Draft: Major library update

theduckylittle:
Bringing GeoMoose into 2022:
- React, ReactDOM updated to V18
- Updated testing library to @testing-library/react (deprecates Enzyme)
- Updates OL to latest.
- Major refactor of the querying code
- Moves to redux-toolkit


klassenjs:
@10f203e2 (just for reference of what code I'm talking about, because I figure this is going to get rebased a lot) 

Generally it looks like it is still working.  The two things I found are also issues in 3.10.

1. I don't like the not highlighting on search parcels is the default now.  It makes it look broken in my eyes. 
2. Geocode Address crashes (pick something outside the demo area)

klassenjs:
@theduckylittle: FYI the tests are failing.  I think it is something related to the test harnesses and maybe them not supporting some of the newer browser API because if all the code was broke that the tests say is broke, nothing would work.

theduckylittle:
@klassenjs Tests are fixed. Had to shim the resizer in the test-rig.

Issue #738, opened 2022-06-30
Search Parcels: highlighting broken


klassenjs:
Results in map don't highlight immediately after search or on hover over the results.

PR #741, opened 2022-07-20
Ensure that highlight defaults to true

theduckylittle:
The parser was treating undefined as `false` which
is the opposite of the desired behavior.

refs: #738


chughes-lincoln:
Interestingly, on my build searching parcels highlights even if you set highlight="false" in the search template. 
Identify, select, and search all correctly default to true, and you can still set identify and select to false if you want.

Issue #750, opened 2022-10-04
Select/Identify Service State Issue


chughes-lincoln:
On the 3.10.2-beta, doing a select, and then an identify, does an identification based on the selection geometry.

![image](https://user-images.githubusercontent.com/43766648/193940428-5e2bbe6a-01a3-4099-80cf-7de334cfe803.png)

chughes-lincoln:
On a related note, if you do a measure (point, line, area), and then click on Identify, it will do an identification based on that measure geometry.

chughes-lincoln:
Fixed by https://github.com/geomoose/gm3/pull/762

PR #762, opened 2022-12-27
Make startService async

theduckylittle:
- Allows for composing the reset actions in the action instead of throughout components/reducers.
- Fixes but where an autoGo service was using the wrong geometry when switching services.
- Consolidates the logic between the application.js API and the toolbar button API.

refs: #750

### DIFF
--- a/src/gm3/actions/query.js
+++ b/src/gm3/actions/query.js
@@ -9,12 +9,23 @@ import { getMapSourceName } from '../util';
 import { changeTool, setSelectionBuffer, clearSelectionFeatures, addSelectionFeature } from './map';
 import { clearFeatures, addFeatures } from './mapSource';
 
-export const startService = createAction('query/start-service', (serviceName, defaultValues = {}) => ({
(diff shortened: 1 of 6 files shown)


chughes-lincoln:
It looks like now if click identify (but don't do an identify on the map), the select tool no longer works. It opens the tool and allows you to create a geometry on the map, but it throws a console error when you click Go.

image

theduckylittle:
I'm not seeing the same behaviour. Can you double check which layer you're selecting from? Using the default (Parcels), I'm able to:
- Click Identify
- Click Select
- Draw a Polygon
- Click "Go"
- I get parcel results on the map.

chughes-lincoln:
It looks like the error only occurs if you don't have a defaultLayer set in the select service in app.js.

theduckylittle:
Not sure if this has been fixed incidentally on not but I'm still unable to re-create that behavior following the steps of just commenting out the `defaultLayer...` line in `app.js`. But I'm also not wanting to block an actual fix (this code) with another issue (a misconfigured service behaving badly).

klassenjs:
Merging based on fixed in later PR.

PR #766, opened 2022-12-31
Adding "Super Search" to the demo

theduckylittle:
This provides a multi-field, multi-operator, multi-modal search to the demo.

Open questions:
- Should we update the base "search" service to accommodate more of this flexibility without stuffing it into `app.js`?
- Should this be in a how-to instead of the code?
- Should this be the default "search"? It's very un-Google.

image


elil:
This provides exciting functionality!  

>  - Should this be the default "search"? It's very un-Google.

No, this search (while very useful and powerful) is not an accessible and popular interface to data.  It starts with "let me learn more about your data structure."  This hits power users and lets them slice and dice but I suspect those users are less than 5% of users.  

> - Should we update the base "search" service to accommodate more of this flexibility without stuffing it into app.js?

If I understand this correctly, I think this is really a different question.  Is this question, "how should more advance, less commonly configured code and functionality be documented and demonstrated"?  I think the default simple search should stay the default in the demo and code, that is what most users will configure and use.  I think that some recent discussion on testing also hit on this concept https://lists.osgeo.org/pipermail/geomoose-users/2022-December/007076.html.  Stuffing less common configurations and testing into a referenced sidecar .js or separate mapbook.xml file might but a good way to do this.  Keeps the demo (code) clean and simple but lets people pursue further things.  There's pluses and minuses to this approach for sure.  

> - Should this be in a how-to instead of the code?

This is another approach and keeps things relatively simple.  This is probably the default approach or the first pass approach and then things could progress from here if warranted.

theduckylittle:
I've gone with the "stuff it in a doc" strategy. Even includes a screenshot!

Issue #375, opened 2019-04-18
Measure feature from result fails when parcel isn't contiguous


chughes-lincoln:
Trying to measure the area of a feature from results fails when the polygon isn't contiguous (in my 3.3.0 setup or the 3.3.1-beta on geomoose.org).

i.e. search for or select parcels, click Measure, Feature from: Results, and click on a parcel that isn't contiguous (i.e. property owned by 'Mitchell Walters')

theduckylittle:
This looks to be an internal OpenLayers bug with the getArea() function. We'll need to research either using turf's implmentation or the ol.sphere version of getArea. I'll leave this for 3.4 but don't want to make another tweak to the measure tools this close to release.

chughes-lincoln:
Just to document this a bit more, it looks like it also fails when a feature has holes (i.e. it doesn't only fail when a feature is a multi-part polygon).

chughes-lincoln:
Tested it out on my site, now works for multi-part features and features with holes.

PR #768, opened 2023-01-02
Switch to spherical area measurement

theduckylittle:
This is a bit more accurate and avoids reprojection to UTM for area measurements.

Also handles graceful degenerate cases.

refs: #375


klassenjs:
For the parcel containing 'Le Nores Airport (historic)' (PIN :130230001001):

Old Area: 3,455,444 sq.ft. (79.29 ac)
New Area: 3,450,843 sq.ft. (79.18 ac)
Identify Acres: 79.40 (presumably calculated according to local survey, but for all I know it is just whatever an old version of some GIS automatically calculated).

So we're looking at about a 0.1%-0.2% difference (at least for the demo).  I'm not sure which one is (closest to) correct.

I'm not 100% convinced either is more accurate.  UTM isn't an equal area projection and has distortion.  But spherical is also wrong.  Both are assuming some height above spheroid/ellipsoid that very likely not the same as the local ground elevation so neither will match direct ground measurements exactly.

In any case we should probably call out this change in the release notes in case anyone was relying on the old area numbers.

elil:
When PostGIS Geography datatypes came out, there were some details on area and distance calculations.  I think it involved moving to an appropriate projection, completing the calculations and then returning - or maybe that was the work around before Geography datatypes?

Size of object measured can impact things too (maybe a state-sized geometry would be different than an 80 acre parcel).  Latitude can have a big impact as well depending on other factors.  I think we have to limit ourselves to GIS and projections, not trying to find grid to ground scale factors to apply.  This is around the corner in low distortion projections (LDPs - often getting that property by being at the right height to get a scale factor of 1).  LDPs are becoming at least available if not yet the norm in the US and some other regions.  Digging back out of the rabbit hole, for where things stand now, I think that the move to ol/sphere is a reasonable approach.

Agree with Jim about release notes.

________________________________

theduckylittle:
Definitely worth adding to the release notes. Might even be worth switching to using the `getLength` version from this package as well to ensure that lines are treated in a consistent manner.  I'll PR that separately when I get a change (hopefully in the next few days).

Issue #776, opened 2023-01-06
Super Search behaves oddly with default values other than 'do not use'


chughes-lincoln:
For the super search the first operator choice for each of the search options is 'do not use'. If you change that to something else (i.e. 'ilike'), it still interprets that operator as 'do not use' unless you manually change it in the drop-down.

For example, let's say I set the operators as 
```
{value: 'ilike', label: 'Similar to'},
{value: 'eq', label: 'Equal to'},
{value: 'in', label: 'In list'}
```
If I don't change anything it will interpret the operator as 'do not use' rather than 'ilike'. If I want an 'ilike' search I would need to manually change to 'eq' or 'in' in the drop-down, and then switch back to 'ilike'.

Perhaps I'm misinterpreting how this search should be used, but I was hoping it would behave more like the old geomoose 2 advanced search, where you would be able to set a default comparitor, and 'do not use' was assumed whenever a user didn't enter something into a given box (not that emulating geomoose 2 is the goal).

chughes-lincoln:
Fixed by https://github.com/geomoose/gm3/pull/762

PR #786, opened 2023-01-11
Fix default value settings for select boxes

theduckylittle:
Fix a long standing bug with <select> boxes not properly setting their default values when an express default value was not set by the administrator.

refs: #776

### DIFF
--- a/src/gm3/components/serviceForm.js
+++ b/src/gm3/components/serviceForm.js
@@ -29,7 +29,7 @@ import { withTranslation } from "react-i18next";
 import TextInput from "./serviceInputs/text";
 import SelectInput from "./serviceInputs/select";
 import LengthInput from "./serviceInputs/length";
-import LayersInput from "./serviceInputs/layersList";
+import LayersInput, { getLayerOptions } from "./serviceInputs/layersList";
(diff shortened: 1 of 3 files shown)


chughes-lincoln:
It looks like this may have introduced a bug with the select tool when layers aren't selectable (perhaps a bit more complicated than that as the OSM layer doesn't cause this issue). If you turn on those layers and open the drop-down for the select tool they show up as empty gaps in the drop-down. 

It also has some odd interactions with the drawing and markup layer. Drawing and Markup now appears in the drop-down. If you add a sketch and then select by drawing and markup it seems to behave as if the sketch tools are active.

klassenjs:
Can confirm what @chughes-lincoln said.

Turn on WX Radar, Go to Select Features, Click the Layers select box.

Also, with this change Drawing and Markup is now in the list.  (I'm not sure this is necessarily a bad thing, but it is likely the result of an unintended side effect).

theduckylittle:
I have a fix I’ll be pushing. I found the bug on the plane but haven’t been
able to push it. 😂

theduckylittle:
Added the bug fix and it's captured here. I rebased and squashed my commits so there's not a unique commit for it.

PR #799, opened 2023-02-10
Ensure measure has a default tool

theduckylittle:
Defaults the measure tool to a polygon.

### DIFF
--- a/src/gm3/components/measure/calc.js
+++ b/src/gm3/components/measure/calc.js
@@ -148,3 +148,27 @@ const PRECISION_LOOKUP = {
 
 export const chompFloat = (value, precision = 2) =>
   Math.floor(value * PRECISION_LOOKUP[precision]) / PRECISION_LOOKUP[precision];
+
+export const normalizeGeometry = (selectedFeatures) => {
+  if (selectedFeatures.length > 1) {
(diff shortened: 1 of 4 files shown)


klassenjs:
Should the default be `line` (so as to not change behavior if the user leaves it un-configured)?  Otherwise looks fine.

elil:
Measure defaulting to line seems reasonable to me
________________________________

klassenjs:
Back to @theduckylittle to merge/edit/whatever.

elil:
Defaulting to other types is fine too if there is a reason for it.  Only mild preference for line since that is what I measure most and maybe previous defaults were.
________________________________

tchaddad:
+1 for defaulting to line so as not to change behaviour

theduckylittle:
Is back to line...

klassenjs:
It looks like the "Fix tests" commit broke tests.

theduckylittle:
Yeah, I was doing some testing in different branches and a local merge got pushed that wasn't complete. Everything is passing now and I've cleaned up the git history.

theduckylittle:
Also, screen shot for posterity of changes. I updated the styling to fit in with more of the application:

image

PR #803, opened 2023-02-13
Fix for "Start over" with query.

theduckylittle:
Clears about both the query-string and the hash string.


chughes-lincoln:
This looks great, only potential issue I've noticed is that Start Over also clears any custom mapbook parameters (i.e. via https://github.com/geomoose/gm3/pull/674/commits/6345cd447dbc047837355a4cd70c3c6755268e2a).

I suppose the only time this would be an issue would be if someone wanted to have the ability to have multiple mapbooks as well as the ability to query by the url using the same app.js.

Issue #811, opened 2023-03-01
Filtering Results from Grid is Broken


chughes-lincoln:
Filtering results throws a console error

chughes-lincoln:
Fixed by https://github.com/geomoose/gm3/pull/814

PR #814, opened 2023-03-14
Fix filter function for grid

theduckylittle:
- Fixes crash with the modals
- Fixes filtering the actual grid content
- Fixes extra "console.log" in the template.


theduckylittle:
refs: #811

Issue #588, opened 2020-08-12
Query results: Move Zoom-to tool to tool bar


brentfraser:
Move the big zoom-to-results "button" to a smaller icon on the results tool bar (along with the clear-results icon).  So instead of:

![image](https://user-images.githubusercontent.com/1638313/90029728-6ceed880-dc78-11ea-9999-8cc19a4227c7.png)

Have:

![image](https://user-images.githubusercontent.com/1638313/90029507-2dc08780-dc78-11ea-9be3-ab9c29c5bbfb.png)

brentfraser:
To move the zoom-to-results tool into the tool bar:

```
diff --git a/src/gm3/components/serviceManager.js b/src/gm3/components/serviceManager.js
index 6b7eaf8..bae20b8 100644
--- a/src/gm3/components/serviceManager.js
+++ b/src/gm3/components/serviceManager.js
@@ -150,13 +150,6 @@ class ServiceManager extends React.Component {
                     <div className='label'>{this.props.t('layers')}</div>
                     <div className='value'>{ layer_count }</div>
                 </div>
-
-                <div className='results-info-item zoomto'>
-                    <div className='label'>{this.props.t('zoomto-results')}</div>
-                    <div className='value' onClick={() => { this.props.zoomToResults(query); }}>
-                        <span className='icon zoomto'></span>
-                    </div>
-                </div>
             </div>
         );

@@ -165,6 +158,13 @@ class ServiceManager extends React.Component {
                 <div className='results-header'>
                     { this.props.t(service_title) }
                     <div className='results-tools'>
+                        <i
+                            title={this.props.t('zoomto-results')}
+                            className='icon zoomto'
+                            onClick={() => {
+                                this.props.zoomToResults(query);
+                            }}>
+                        </i>
                         <i
                             title={this.props.t('results-clear')}
                             className='icon clear'

```
and
```

diff --git a/src/less/results.less b/src/less/results.less
index 6d2cdcc..8596224 100644
--- a/src/less/results.less
+++ b/src/less/results.less
@@ -86,7 +86,7 @@
         display: flex;

         .results-info-item {
-            width: 30%;
+            width: 50%;

             padding: .25em;
             margin: 5px;

```

tchaddad:
I don't necessarily have a strong opinion on this, but this change would seem like it might be unfriendly to folks using the UI on a mobile device, if that matters to you...

brentfraser:
[...]

PR #820, opened 2023-03-21
Refactor of results summary

theduckylittle:
- Moves counts into a small table below the actual controls
- Makes the delete/clear control larger
- Changes the filter results message.


theduckylittle:
refs: #588

chughes-lincoln:
It would be nice if the boxes re-sized themselves to use all the space like in Brent's example.

elil:
I'm checking if I know what these means:
Features: 395
Filtered 457 features from the results

Does this mean that 852 (395+457) features were selected or returned from a search?  And then some user interaction filtered 457 features out of those 852 results in the table leaving only 395 features in the table?

chughes-lincoln:
Yes, the search or selection returns 852 results. Then filtering in the grid removes 457 results leaving 395.

klassenjs:
The wording seems non-obvious to me.  Filtered to me implies the number of results left after filtering rather than the number of results filtered out.   To me it would be more natural to list the total number of results, and the number of results shown.

chughes-lincoln:
On a related note, should buffering results only be run on the results remaining after a filter?

Currently the buffer tool buffers all results (including results removed by a filter), but it uses the count of the remaining results to determine whether a buffer can be run (you can't buffer more than 100 results at a time).

i.e. if I try to buffer 500 results, it throws an warning telling me I can't buffer that many results. If I filter results to get below 100 results, it allows me to buffer, but it still uses all 500 results.

chughes-lincoln:
@tchaddad I like the wording, it might help to also say 'X results are filtered out' instead of just filtered.

PR #821, opened 2023-03-21
Update mapbook.xml

brentfraser:
fix runways map-source name to correspond to content of app.js

### DIFF
--- a/examples/desktop/mapbook.xml
+++ b/examples/desktop/mapbook.xml
@@ -287,7 +287,7 @@
         </layer>
     </map-source>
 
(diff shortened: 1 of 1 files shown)


klassenjs:
The other mapbooks (in desktop and mobile) also need to be updated.  And app.js likely is also out of date between desktop and mobile.   However, as there was an existing mess before this PR, it might be cleaner to apply this as is and then have another PR to sync things up.  Thoughts?

brentfraser:
I agree; it would be best to do a separate Issue+PR to clean up the mapbooks and app.js.
This was just a very small fix to get searching runways working.

Issue #819, opened 2023-03-20
Print Legend Adds Blank Spaces


chughes-lincoln:
There are a few situations that result in blank spaces between layers in print legends, which can make the legend look a bit less professional and can make it more difficult to fit all of your layers in the legend. I'm not sure how much of this is a mapserver vs a geomoose issue, but I thought I'd document them.

1) If a layer is on in the catalog but not visible at that scale.

2) If you have multiple layers grouped together into a single layer in the catalog. If the legend is on for that catalog layer, any layers that don't have a class name will show up as blank spaces. The issue here is that you can't turn off legends for individual map-sources, so the only way to avoid adding legends for individual map-sources is to omit the class name.

3) If you have a layer that has multiple classes in the map-source, there will be a gap in the legend after the layer, which is proportional to the number of classes in the layer.

Example below:
![image](https://user-images.githubusercontent.com/43766648/226475901-097ee92f-6ca4-410b-9fe4-dd7f333efb93.png)

theduckylittle:
Created a filter to catch the "placeholder" images that Mapserver returns.

chughes-lincoln:
This fixes 1 and 2. I'm still seeing a larger gap after a layer if the layer has multiple classes. That's a weird one and only really an issue if someone has a complex dataset with a lot of classes. Fixing 1 and 2 is a huge improvement.

![image](https://user-images.githubusercontent.com/43766648/234374899-a03339f9-fb7a-4290-b635-2f1687b86c4f.png)

theduckylittle:
This is going to be mapserver... I didn't see any easy fixes for this in the documentation.

theduckylittle:
... the only way to fix this on the client would be to make some big assumptions about the legends:
1. That they are coming from mapserver and generally follow that format and/or...
2. That if a group of N number of lines in the image is all the same that means it's "empty" and the "empty" lines should be removed.

I'm not sure I'm wholly comfortable making those assumptions.

chughes-lincoln:
I kind of figured that but was still holding out hope. I'm comfortable enough with the enhancements that have already been made.

PR #826, opened 2023-04-25
Clean up empty legends

theduckylittle:
- If a legend is less than 8 pixels tall, its assumed to be a space holder.
- Check if the image is all the same color, if so, consider it empty as well and do not include it in the print.

refs: #819


klassenjs:
I'm getting a `Uncaught (in promise) ReferenceError: OffscreenCanvas is not defined` in Firefox 102 after I click the print button in the print-preview dialog.

Issue #810, opened 2023-03-01
Print - Image in Template - i.e. north arrows/organization logo


chughes-lincoln:
Adding an image in a print template (https://docs.geomoose.org/3.x/howto/print-layouts.html?highlight=print) no longer works (at least I haven't figured out how to get it working). Previously you could convert your image to a base64 string and reference it in the image_data parameter. On 3.10.0 you can also just directly reference a png.

theduckylittle:
Ack! The new lint fixes got a little aggressive. New MR handles it properly.

chughes-lincoln:
Fixed by https://github.com/geomoose/gm3/pull/827

PR #827, opened 2023-04-25
Fix broken `image_data`

theduckylittle:
Fixing camel case variables got aggressive. This rights that wrong in the `image_data` case for print templates.

refs: #810


klassenjs:
I don't have a test case for this one, @chughes-lincoln can you verify this works?

chughes-lincoln:
Confirmed base64 and png's now work again. Thanks!

It may be a good idea to add a png of the GeoMoose logo to one of the demo templates. Partly to make it easier to test in the future and partly to give people an example of including an image.

Issue #828, opened 2023-04-25
+/- Zoom Buttons zoom to whole world


klassenjs:
This might be related to using bbox on the hashtracker.  I haven't really tracked this down yet, but noticed it when testing the zoom creep pr.

theduckylittle:
This is a regression due to how we handle zoom-by-zoom and zoom-by-resolution. I'm digging in now.

chughes-lincoln:
Not sure if there were any particular steps needed to replicate this, but I rebuilt on the demo, tested the default xyz, lonlat, and bbox and didn't see any issues.

PR #829, opened 2023-04-25
Fix bug resolving resolution to zoom in map controls

theduckylittle:
Resolution is the authoritative variable for
the map "zoom," this was assuming that zoom was always set as well.

This will use zoom, if defined, but otherwise calculate the appropriate next step from the resolution.


theduckylittle:
refs: #828

PR #837, opened 2023-05-30
Measure Tool Enhancements

theduckylittle:
The goal of this PR is to provide:

- [X] Measure features that are preserved on the map.
- [x] Measure features with a new, more complex styling.
- ~~[ ] Hover interactions between the features on the map and the features in the measure list.~~ Went with better cartography instead of the hover interaction.

### DIFF
--- a/examples/desktop/mapbook.xml
+++ b/examples/desktop/mapbook.xml
@@ -82,7 +82,7 @@
     </map-source>
 
     <map-source name="vector-parcels" type="mapserver-wfs">
-        <file>demo/parcels/parcels.map</file>
+        <file>./demo/parcels/parcels.map</file>
         <param name="typename" value="ms:parcels"/>
         <config name="pixel-tolerance" value="0"/>
         <transform attribute="EMV_TOTAL" function="number"/>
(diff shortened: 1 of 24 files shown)


chughes-lincoln:
This looks great so far!

I noticed a few things:

- Measure points no longer show up on the map - this might be included in your new complex styling
- Measure lines - subsequent measurements recalculate the final segment of previous measurements 
- Draw Polygon throws a console error - polyline attribute points: Expected number - I'm not sure what impact this has other than 
- I'm not sure how useful the incremental area calculation is. I could see why people might want it, but it might be confusing for the average user. I think having results for each individual line segment is intuitive (as well as the sum), but having the history of all potential versions of a polygon isn't as intuitive. I feel like when someone's measuring lines, they want the distances and bearings of each line segment, but when people are measuring polygons, they're goal is measuring the final polygon. I do really like having a picture of the polygon in the measure results.
- If I start a measurement, and then close the measure tool before finishing the measurement, the result will remain in the results tab and can't be cleared (unless I start another tool or start another measurement)

Issue #835, opened 2023-05-25
Demo use of Min/Max filter is non-functional


tchaddad:
In the main demo there is a filter that uses Min and Max values that is non functional:
![image](https://github.com/geomoose/gm3/assets/1042618/e50b0291-3367-4fd9-a6de-681327338f06)

The failure is showing up in the Console as:

```
Uncaught TypeError: this.state.value is undefined
    close filter-modal.js:78
    onClick modal.js:75
    onClick modal.js:39
    React 15
        callCallback
        invokeGuardedCallbackDev
        invokeGuardedCallback
        invokeGuardedCallbackAndCatchFirstError
        executeDispatch
        processDispatchQueueItemsInOrder
        processDispatchQueue
        dispatchEventsForPlugins
        dispatchEventForPluginEventSystem
        batchedUpdates$1
        batchedUpdates
        dispatchEventForPluginEventSystem
        dispatchEventWithEnableCapturePhaseSelectiveHydrationWithoutDiscreteEventReplay
        dispatchEvent
        dispatchDiscreteEvent
```

PR #838, opened 2023-06-17
Fix for range filter modals

theduckylittle:
Update to use `this.state.value` instead of the previous `min` and `max`.

refs: #835


chughes-lincoln:
This fixed the issue on my build of the demo:

image

PR #910, opened 2025-07-18
Update webpack-dev-server and jspdf

klassenjs:


klassenjs:
Still need to fix webpack dev server options.

PR #923, opened 2025-09-19
Draft: Add GeoParquet support to the mapsource types

theduckylittle:
This is a fairly raw version of the parquet support. It loads the entire parquet file, on the main thread.

Missing from the commit is the demo parcels file.


klassenjs:
See #938 for future GeoParquet development.

Issue #930, opened 2025-12-13
Catalog group chevrons not rendering in iOS/Mac


chughes-lincoln:
I noticed the chevrons for opening and closing the groups in the catalog tab weren't rendering on my Mac or iPhone, independent of the browser I was using. Noticed no issues in Windows using Edge, Chrome, or Firefox.

Mac, Chrome:
<img width="351" height="675" alt="Image" src="https://github.com/user-attachments/assets/c5b2885a-7410-4a07-a4cb-92b088cda168" />

Mac, Firefox:
<img width="350" height="622" alt="Image" src="https://github.com/user-attachments/assets/e985e33c-20ce-46b9-9fb3-247e2fc36305" />

iOS, main demo, Firefox:
<img width="603" height="1311" alt="Image" src="https://github.com/user-attachments/assets/fb4dd2b2-f0f3-41fb-92de-8f88d54e59b4" />

iOS, lite demo, Firefox:
<img width="603" height="1311" alt="Image" src="https://github.com/user-attachments/assets/10a9d8f4-bfd1-4f7c-b370-7d198743fa14" />

It looks like it doesn't recognize the character code. On a whim I changed it to a different arrow I found and it rendered correctly. I found the character we're currently on this site - https://www.compart.com/en/unicode/U+2B9E - noticeably, even here, it renders fine on windows but shows up as a box on mac.
<img width="1470" height="799" alt="Image" src="https://github.com/user-attachments/assets/853a258c-9305-4848-a8b6-a5835aed36f3" />

Other example of a different code that works:
<img width="1005" height="673" alt="Image" src="https://github.com/user-attachments/assets/c45829db-b0d1-49b6-b2ef-582f074db909" />

chughes-lincoln:
Fixed by https://github.com/geomoose/gm3/pull/931

Issue #925, opened 2025-10-09
Catalog - Open/Close Group Behavior Change


chughes-lincoln:
Ran into a minor issue on 3.14. 

When you hover over a group in the catalog, you get a gray highlight for the entire group-label div. On 3.13 you could click anywhere in the gray highlight to open or close group. On 3.14, you specifically need to click on the group icon, or the text span to open or close the group. If you click between the two or after the text (while still clicking on the gray highlight), nothing happens (which I think is unintuitive to the user - you would think the gray highlight indicates you can click on it to do things - I kept unintentionally clicking between the icon and the text and having nothing happen).

I was able to fix it in the css (essentially, I set the span width for the group-label to 100% so it extends to the end of the tab, and adjusted the padding and margins of the icon so that there isn't a gap between the icon and the text), but I wanted to raise it as an issue for other people.

chughes-lincoln:
Fixed by https://github.com/geomoose/gm3/pull/931

PR #931, opened 2025-12-14
Fix for catalog group icons and title clicks

theduckylittle:
1. Moves the `onClick` event to the entire catalog title `<div/>`
2. Changes to for Font-Awesome icon to prevent font differences amount platforms.

refs: #930, #925


chughes-lincoln (approved):
Everything looks good on my build of the demo

PR #938, opened 2026-02-15
feature: Adding GeoParquet Support

theduckylittle:
Utilizes HyParquet to add GeoParquet support to GeoMoose v3.x. The goals for the PR are:

- [x] Provide Rendering for GeoParquet files with Vector Styling
- [x] Provide support for Identify
- [x] Provide support for Select
- [x] Provide support for Search

The goal is to offer an alternative to WFS or MapServer for rendering and querying for the Parcels dataset.


klassenjs:
* Search Parcels -> Name: "Smith" returns a bunch of roads (these have `null` names).  Single Field Search works as expected.
* Search Parcels returns nothing if the GeoParquet layer hasn't been enabled yet. (not sure if this is new or if this just didn't happen in the demo because the parcels layer is on by default).

klassenjs:
I'm getting "failed to find geometry column" on one of the older geoparquet files I am testing (the metro parcels).  I'm not sure that's a fully compliant file, but I note a difference of the geometry column being `geometry` instead of `geom`.  It looks like the code checks for this though, so I'm confused.

klassenjs:
Testing f0e1d50:

```bash
# GDAL recommended v1.1
gdal vector convert -f Parquet \
        --lco COMPRESSION=ZSTD --lco COMPRESSION_LEVEL=22 \
        --lco GEOMETRY_ENCODING=GEOARROW \
        --lco SORT_BY_BBOX=YES \
        --lco GEOMETRY_NAME=geom \
        Parcels2023.geoparquet Parcels2023-1.1.geoparquet
# -> 184MB, fails to load with exception about parsing geometries

# https://www.geomermaids.com/cookbook/geoparquet/ recommended v1.1
gdal vector convert -f Parquet \
        --lco COMPRESSION=ZSTD \
        --lco ROW_GROUP_SIZE=100000 \
        --lco GEOMETRY_ENCODING=WKB \
        --lco SORT_BY_BBOX=YES \
        --lco WRITE_COVERING_BBOX=AUTO \
        --lco GEOMETRY_NAME=geom \
        Parcels2023.geoparquet Parcels2023-1.1gm.geoparquet
# -> 192MB, works (with smaller input file, I get an OOM on the full file)

# GeoParquet v2.0
gdal vector convert -f Parquet \
        --lco COMPRESSION=ZSTD --lco COMPRESSION_LEVEL=22 \
        --lco ROW_GROUP_SIZE=100000 \
        --lco USE_PARQUET_GEO_TYPES=YES \
        --lco SORT_BY_BBOX=YES \
        --lco GEOMETRY_NAME=geom \
        Parcels2023.geoparquet Parcels2023-2.0.geoparquet
# -> 159MB, works
```

Note: The `GEOMETRY_NAME=geom` is needed otherwise the "Failed to find geometry column in : ..." error occurs.

Note: If loading the entire metro area parcels (1127614 features), I get a "Unhandled out of memory exception".  For the working tests, I limited the source to the first 1000 features.

Update: The full v2.0 file loads without an OOM.  The full 1.1gm file triggers the OOM.  And by loads, I mean eventually; loading the full metro isn't a practical user experience at the moment.

klassenjs:
Hopefully, the geometry column detection can get fixed.  GDAL assumes `geometry,wkb_geometry,wkt_geometry` are possible geometry column names and defaults to `geometry` when creating files.

The rest of my above comment is just observations and the start of documentation on how to create files and reasonable data sizes.

klassenjs:
Testing: 5584808.  The geometry column name issue is resolved.  The `--lco GEOMETRY_NAME=geom` parameter is no longer necessary.

Case 1.1 from above now fails with `can't access property "name", u[0] is undefined`.  I think it is reasonable to not support the v1.1 GeoArrow geom format as long as we document that.  From what I can tell, it was an interim format that had introduced some limitations and the v2.0 format has all of the advantages and none of the limitations of this format.  v1.1 WKB and v2.0 are most likely the formats found in the wild.

Case 1.1gm (WKB) and 2.0 work.  I'd expect v2.0 to be the preferred format going forward.

So, I believe this is now the preferred GDAL (v3.12+) command to generate GeoParquet for GeoMoose:
```
gdal vector convert -f Parquet \
        --lco COMPRESSION=ZSTD --lco COMPRESSION_LEVEL=22 \
        --lco ROW_GROUP_SIZE=100000 \
        --lco SORT_BY_BBOX=YES \
        --lco USE_PARQUET_GEO_TYPES=YES \
        Parcels2023.gpkg Parcels2023.geoparquet
```

chughes-lincoln:
Finally getting around to testing this

Some comments

- It looks like the geoparquet file has to be in 4326
- All the services I tested were working - simple search, multi-field search, search-by-list, select
- Interesting speed behavior
  - Searches seemed slower, particularly if you have a bunch of features in different areas; starting a service from a url was slower (likely because it's loading the geoparquet file)
  - Displaying a large dataset was slower - 10 seconds for 45,000 parcels vs 2 seconds for postgres
  - Selections seemed to return results almost instantly, as long as you weren't selecting thousands of parcels
 
- The Jumpy GeoMoose spinner didn't always work - seemed to show on some slow queries - occasionally noticed a slow query where it didn't appear
- Null results come through as null rather than ??? (which I prefer, though I would still prefer it come through as empty). You can still use |blank>N/A to divert null to N/A (or even nothing), but it would be nice to not have to do that for every field in every template
- Is there a way of connecting to a file in a non-web accessible location like you can in a mapfile? Using the URL approach allows users to download the geoparquet file if they know the URL, which could be a dealbreaker for some admins, though I suppose if that were an issue they could probably just restrict access to the directory. I could also see some confusion stemming from people putting the data in the same spot as their mapfiles and being confused when it doesn't work

chughes-lincoln:
The changes to the title/padding look good to me.

PR #961, opened 2026-03-26
chore: 2026-03 Cleanup Dependencies

theduckylittle:
- Fix dev versus application dependencies
  - Drop "vintage" url package for built-in browser capabilities


theduckylittle:
So weird. This was working when testing locally then apparently stopped and it should never have worked in the first place. :shrug: 

Fixed now

PR #977, opened 2026-04-17
Switch GitHub actions to use pinned SHA

klassenjs:
Branches (and even tags) can change to run modified code that can leak credentials and API keys.  This is apparently happening in the wild.

https://github.blog/changelog/2025-08-15-github-actions-policy-now-supports-blocking-and-sha-pinning-actions/#enforce-sha-pinning

### DIFF
--- a/.github/workflows/check-crlf.yml
+++ b/.github/workflows/check-crlf.yml
@@ -22,6 +22,6 @@ jobs:
         uses: actions/checkout@v6
 
       - name: Use action to check for CRLF endings
-        uses: erclu/check-crlf@master
+        uses: erclu/check-crlf@c77fda0a97ea8944aba5f72796b9e937e7d1611c # master
         with: # ignore directories containing *.pdf and *.tab
           exclude: 
--- a/.github/workflows/irc_notify.yml
+++ b/.github/workflows/irc_notify.yml
@@ -16,7 +16,7 @@ jobs:
     if: github.repository == 'geomoose/gm3'
     steps:
       - name: irc push
-        uses: rectalogic/notify-irc@v2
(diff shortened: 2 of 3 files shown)


klassenjs:
I'm not sure if it is best to pin all SHAs or just "third-party" actions.  I'm not 100% on which actions are "third-party", I'm assuming anything that doesn't start with `actions/`

Pinning SHAs has the disadvantage of not taking advantage of security updates automatically.  But it also more consistent/reproducible and doesn't automatically pull in new security vulnerabilities.

We aren't currently doing much that is explicitly security sensitive within the workflows, but I presume that various github tokens get automatically pulled in and so the workflows, if compromised, could use that to do bad things.

We should also look into updating the workflows to current versions.

We use:
```
check-crlf.yml: uses: actions/checkout@v6
check-crlf.yml: uses: erclu/check-crlf@master
irc_notify.yml: uses: rectalogic/notify-irc@v2
tests.yml: uses: actions/checkout@v2
tests.yml: uses: actions/setup-node@v1
```

Current versions are:
`actions/checkout` is at `v6.0.2`
`actions/setup-node` is at `v6.3.0`
`erclu/check-crlf` is at `master`, currently 9mo old.
`rectalogic/notify-irc` is at `v2` which is the latest and 5 years old.

klassenjs:
There are also config options under the repo Settings page we can set to limit what actions can do (no create/approve PRs, make the repo read only, etc.)

klassenjs:
On the organization level, we have Actions permissions set to read-only/no PRs by default, so having no explicit permissions block in the actions.yml files should do the right thing.

theduckylittle (approved):

PR #985, opened 2026-05-25
fix: Drop use of vintage uuid module

theduckylittle:
Migrates to the built in `crypto.randomUUID` function.

This also adds a `@gm3` prefix for easier package organization.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -7,4 +7,7 @@ module.exports = {
     moduleDirectories: ["./src", "./node_modules"],
     setupFiles: ["./tests/setup.js"],
     testEnvironment: "jsdom",
+    moduleNameMapper: {
+      "^@gm3/(.*)$": "<rootDir>/src/gm3/$1",
+    },
 };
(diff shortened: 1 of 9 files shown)


chughes-lincoln:
Am I missing something with how to incorporate these changes? When I rebuild on the 3.15.2-beta I get a couple errors and the map/catalog/toolbar fail to load.

image

I tested on my version of the demo on Windows, I replaced everything in the geomoose folder and everything in the desktop folder (though I don't see anything that would require configuration changes). I get the same results when building myself and when downloading the nightly development build for MS4W.

klassenjs:
@chughes-lincoln What browser?  `crypto.randomUUID()` is supposed to be provided by the browser.

chughes-lincoln:
I was testing on Edge (replicated on firefox and chrome). Upon googling it looks like crypto.randomUUID requires https (which I don't have on my dev site). It works via localhost.

klassenjs:
Well, that's interesting.

@theduckylittle Thoughts?

theduckylittle:
:thinking: This was working fine for me on non-HTTP. I'll follow up with a new PR that addresses this.

PR #986, opened 2026-05-25
feature: Add support for cloud optimized GeoTIFFs

theduckylittle:
Add an example for direct access to a COG from a bucket.

No need for a raster server, just go for it!

Updates OpenLayers from 6.x to 10.


klassenjs:
The build works.  The tests fail.  The jester has no sense of humor.

klassenjs:
I think this is the fix for the tests, but I'm not 100% the implication of these changes, so leaving as a comment instead of a commit:

```patch
diff --git a/jest.config.js b/jest.config.js
index a463195..377465f 100644
--- a/jest.config.js
+++ b/jest.config.js
@@ -3,7 +3,7 @@ const {defaults} = require('jest-config');
 module.exports = {
     ...defaults,
     transform: {"\\.js$": "babel-jest"},
-    transformIgnorePatterns: ["/node_modules/(?!(ol|ol-mapbox-style|jsts|geotiff|quick-lru|rbush|quickselect|pbf)/)"],
+    transformIgnorePatterns: ["/node_modules/(?!(ol|ol-mapbox-style|jsts|usng|usng-map-collar|geotiff|quick-lru|rbush|quickselect|pbf)/)"],
     moduleDirectories: ["./src", "./node_modules"],
     setupFiles: ["./tests/setup.js"],
     testEnvironment: "jsdom",
diff --git a/webpack.config.js b/webpack.config.js
index 5d1744c..768550d 100644
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -70,7 +70,7 @@ module.exports = env => {
                     path.join(__dirname, 'node_modules/'),
                 ],
                 exclude: function(absPath) {
-                    var acceptable = ['ol', 'mapbox-to-ol-style', '@mapbox', 'jsts'];
+                    var acceptable = ['ol', 'mapbox-to-ol-style', '@mapbox', 'jsts', 'usng-map-collar'];
                     if(absPath.indexOf('node_modules') < 0) {
                         return false;
```

klassenjs (approved):

PR #989, opened 2026-06-03
fix: Drop use of crypto.randomUUID

theduckylittle:
This fixes ID generation for HTTP hosted sites (common for internal deployments)

@chughes-lincoln -- see if this fixes your problem!


chughes-lincoln:
Yep, works now over HTTP

chughes-lincoln:
This also works for me over http

PR #990, opened 2026-06-03
fix: Broken bookmarks

theduckylittle:
Fixes the bookmark regression from the fix of the custom URL parser.


chughes-lincoln:
This works for me

PR #992, opened 2026-06-05
feature: Add Annotations to Measurements

theduckylittle:
Adds new annotations for measurements:

image

Which can also be toggled off with the new labels toggle:

image


chughes-lincoln:
This looks great and I think some of the people that use our site will be really excited by this!

I have some thoughts on how the measure units change for area when you switch between length and area. I think it makes sense that changing length units from feet to meters will change area units from sq. feet to sq. meters. I think it's less obvious for chains (and rods - which isn't in the demo but is an option in the underlying code). I was talking with the surveyors in our office about what would be intuitive to them, as they're the ones that tend to use chains and rods, and they recommended Acres, over square feet (which is what chains defaults to) and square rods (which rods defaults to - though I'm not sure if square rods can be set in the configuration, so it looks like the user has no area unit selected).

My only other comment is if the configuration for turning the units off by default should be with the other measure configurations:

```
var app = new gm3.Application({
        mapserver_url: CONFIG.mapserver_url,
        mapfile_root: CONFIG.mapfile_root,
        map: {
             measure: {
                    areaUnits: ["ft", "mi", "a", "h", "m", "km"],
                    lengthUnits: ["ft", "mi", "ch", "r", "m", "km"],
                    showMeasureLabels: false
          }
    });
```

chughes-lincoln:
I did notice some interesting behavior when selecting from results. Labeling the area is useful, but I could see people making false assumptions about the labels for the line segments, as it can be hard to tell when a label isn't being displayed because it doesn't fit (particularly when you have polygons with vertices without obvious changes in direction).

Take this parcel for example, the section along SE Whalesong Dr shows as 127 feet, but you would need to zoom in to see a bunch of sub-1ft segments to see that it's actually closer to 230ft

image

Or this example where a couple of line segments aren't showing:

image

It might be useful to disable labeling line segments for features from the results (or at least letting the admin turn them off for results). I may also be overanalyzing things. I like the labeling for when users are adding their own lines and polygons.

theduckylittle:
@chughes-lincoln , just pushed some changes based on your earlier feedback.

We could turn off label suppression? That would ensure all the little bits are coming through but would be, ultimately, much uglier when it's dense.

chughes-lincoln:
I think those changes are good.

Overall I like how the labels are suppressed for user-added lines/polygons more than I dislike how they're suppressed for features from results. People capable enough to measure features from results probably know enough to zoom in to see all the labels.

tchaddad:
Not sure if I followed Chris' last comment, but I wanted to agree that users will really like this feature! 

I think that's particularly true for things that users draw on the map, and think this feature can be used in conjunction with regular sketches in interesting ways.

klassenjs:
Is this one ready to merge or still under discussion?

chughes-lincoln:
Looks good to me

theduckylittle:
A little cowboy here but I'm merging this with lack of controversy (it's also a dependency for some new work coming in...)

PR #996, opened 2026-06-09
fix: Font structure, radio and checkbox rendering

theduckylittle:
I noticed some bugs on HiDPI screens and definitely some quirks with some of the recent font-changes on `main`.

Spent some time refactoring and then massaging the fonts a little to fix:
1. Issues with font-sizes being arbitrary (they're not enumed in a single variables file)
2. Issues with the radio buttons and checkboxes. @klassenjs noticed this too, where on some screens the font rendering would make these look uneven. That's been corrected now.

image


klassenjs:
Before:
image

After:
image

PR #999, opened 2026-06-14
feature: Add print to scale

theduckylittle:
Adds new scale options to printing. This trades off the with fit-to-box model from before.

This diff did get rather large but it was necessary to fix some *old* assumptions built into the map component.

image

### DIFF
--- a/docs/howto/configure-print-scales.rst
+++ b/docs/howto/configure-print-scales.rst
@@ -1,0 +1,84 @@
+Configuring Print Scales
+========================
+
+The fixed scales offered in the print dialog can be set via the
+application configuration's ``print.scales``. The "Scale to fit" options
+are always available; when ``print.scales`` is omitted a built-in set of
+common scales is used.
+
+The configuration is passed to the ``gm3.Application`` constructor in your
+``app.js``.
+
+::
+
+    var app = new gm3.Application({
+        mapserver_url: CONFIG.mapserver_url,
+        mapfile_root: CONFIG.mapfile_root,
+        print: {
+            scales: [
+                {scale: 50},
+                {scale: 1000},
+                {scale: 24000},
+                {label: "1000 ft / in", metersPerPt: 4.2333}
+            ]
+        }
+    });
+
+To offer only the "Scale to fit" options and no fixed scales, set
+``scales`` to an empty array:
+
+::
+
+    var app = new gm3.Application({
+        mapserver_url: CONFIG.mapserver_url,
+        mapfile_root: CONFIG.mapfile_root,
+        print: {
+            scales: []
+        }
+    });
+
+Each entry sets the scale one of two ways:
+
+* ``scale`` - A ``1:N`` ratio (e.g. ``24000`` for ``1:24000``). Simplest,
+  and the natural fit for metric printing.
+* ``metersPerPt`` - The ground distance, in meters, that one PDF point of
+  paper represents (72 points per inch). For "per inch" scales, divide the
+  per-inch distance by 72: ``1000 ft / in`` is ``1000 * 0.3048 / 72``.
+
+``label`` is optional; a ``1:N`` label is derived from the scale when
+omitted.
+
+The built-in scales
+-------------------
+
+When ``print.scales`` is omitted, the engineering scales below are
+offered. They cover the US local-government use cases discussed in
+`geomoose/gm3#999 <https://github.com/geomoose/gm3/pull/999>`_. Set
+``print.scales`` to replace the list outright with your own -- for
+metric printing, or to reach the regional scales (1 mile per inch,
+1:100000) that the defaults stop short of.
+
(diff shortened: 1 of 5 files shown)


chughes-lincoln:
People can probably assume it by context, but it might be useful to mention in the doc that the configuration is done in the app.js file.

Also, is there a way to configure it to just use the Scale to fit/higher/highest settings (i.e. just use the Fit presets)? The only way I see to remove the default presets is to add additional presets.

theduckylittle:
> People can probably assume it by context, but it might be useful to mention in the doc that the configuration is done in the app.js file.
> 
> Also, is there a way to configure it to just use the Scale to fit/higher/highest settings (i.e. just use the Fit presets)? The only way I see to remove the default presets is to add additional presets.

Addressed both of these:

1. Docs are more clear
2. Fixed the bug where `[]` meant defaults instead of "no additional scale settings", oversight on my part. I tried to clean up some review warnings that said this mixed types, but then it removed the ability to totally remove print to scale.

klassenjs:
I worry about making the default list too long, but having the full 100, 200, 300, 400, 500, 600 ft = 1in series and 1:6000 would be nice.  Thoughts?

chughes-lincoln:
The empty array worked for me. I'm indifferent about adding more options, but if we were concerned about the list being too long, we could add them to the example desktop configuration (where it's easier to just remove an entry you don't want) or to the doc.

tchaddad:
From the PSC discussion today:

List of possible scale defaults (US Local Govt use cases):

- **1 inch = 20 feet (1:240)** For detailed intersection layouts, utility alignments, and site-specific public works.
- **1 inch = 40 feet (1:480)** Common standard for residential streets and general utility plans.
- **1 inch = 50 feet (1:600)** Frequently used for grading plans, larger parcels, and overall subdivision layouts.
- **1 inch = 100 feet (1:1200)** City-wide overviews, long linear right-of-way maps, or large-scale utility networks.
- **1 inch = 500 feet (1:6000)** Jim's fire example: common in USNG and fire maps (a 1km square fits on a 8.5x11 sheet).
- **1 inch = 2000 feet (1:24000)** Standard scale for 7.5 minute USGS topo quads

chughes-lincoln:
I had a request a while ago for 1 inch = 10 feet (for septic evaluations and permits). I heard back from some Planners and they mentioned 1 inch = 200 ft, and 1 inch = 50ft.

theduckylittle:
Updated with new scales and noted docs!

chughes-lincoln:
Something that might be worth mentioning in the docs - if the admin has configured zoom restrictions, they need to make sure that they adjust the print to scale options so that they're compatible with their zoom restrictions.
- i.e. if I've set a maxZoom of 20, and I update to the geomoose version that contains this PR without changing any of the default settings, users will be able to print to 10ft/in. It will still print with a label saying 1 inch = 10 feet, but in this case the result is closer to 1 inch = 75 feet.
image

I also heard feedback about wanting to be able to export a pdf at a particular scale, while also omitting the 1 inch = 10ft label. The reason being to avoid any complaints/liability for when people take a pdf exported at a particular scale, and then print it at a different one (for example I printed something at Letter - Portrait, but the printer settings defaulted to 11x17 Portrait with Fit to printable area checked - so the 10ft scalebar was still correct because it didn't change the aspect ratio, but it was no longer 10ft/inch).

klassenjs:
> The reason being to avoid any complaints/liability for when people take a pdf exported at a particular scale, and then print it at a different one

Something that says something like "the scalebar should measure exactly 1 inch when printed at the intended scale" can help with this.  Using a 2D reference feature (a square somewhere or maybe even the map frame) could also show issues with aspect ratio.  But I could see where people might want to just turn that off too (or say not to any particular scale).

theduckylittle:
This merged while @chughes-lincoln's and @klassenjs's last comments were still open, and a merged PR can't be reopened — so I've carried them forward to #1021.

What's in it:

- **The scale label could state a scale the map didn't render at.** @chughes-lincoln's `maxZoom: 20` case, where 10 ft/in prints nearer 75 ft/in but is still captioned "1 inch = 10 feet". I treated this as a fix rather than the docs note asked for: GeoMoose now checks what OpenLayers will actually clamp to and omits the caption when it disagrees, warning the admin on the console. The scale bar was already computed from the rendered view and is unchanged.
- **`scaleLine.printScaleLabel: false`** — print to scale, keep the bar, drop the caption. For the reprint/liability case.
- **Docs** for both, in `configure-print-scales.rst` and `configure-scaleline.rst`.

Left open for discussion over there rather than guessed at: @klassenjs's "the scalebar should measure exactly 1 inch when printed at the intended scale" wording, and the 2D reference feature for spotting aspect-ratio distortion. Both seem right to me; I'd just rather settle the shape with you first.

Sorry for the premature merge, and thanks to you both for catching the scale-label problem before it reached anyone's plat.

PR #1000, opened 2026-06-16
feature: Update the scale line to be more clear

theduckylittle:
Make the scale line easier to read on the PDF map.

Uses more traditional "line with ticks" and adds a contrasting background with nice rounded corners.

[print_1781620271756.pdf](https://github.com/user-attachments/files/29007521/print_1781620271756.pdf)


chughes-lincoln:
This looks good.

No background:
image

Aerial:
image

Hillshade:
image

PR #1013, opened 2026-07-31
feature: Feature Report!

theduckylittle:
Adding the feature report functionality from GM2.

This depends on other branches. Attached a new sample output and screen shot!

image

[Uploading print_1785455926994.pdf…]()


chughes-lincoln:
A couple thoughts
- Maybe this is just a configuration issue, but there's still an option for setting the map title in the print modal, even though that's overwritten in the report json
- By default it zooms right up to the parcel boundary. It would be nice if it would be slightly zoomed out so that you could see the outline (or maybe default to higher/highest resolution scale to fit which zooms out)

image

theduckylittle:
> A couple thoughts
> 
>     * Maybe this is just a configuration issue, but there's still an option for setting the map title in the print modal, even though that's overwritten in the report json
> 
>     * By default it zooms right up to the parcel boundary. It would be nice if it would be slightly zoomed out so that you could see the outline (or maybe default to higher/highest resolution scale to fit which zooms out)
> 
> 

I've added a few things to address this:

1. The map-title can now be overridden or not allowed. This is done based on the feature-report template settings.
2. There are now more zoom controls including:
   A. A new `maxScale` parameter.
   B. A new `padding` parameter that allows specifying a nice pixel-unit buffer around the extent of the feature.

PR #1016, opened 2026-08-02
chore: Add a contributing guideline / LLM policy

theduckylittle:
This is a first cut at an LLM policy for GeoMoose.


klassenjs:
Should we be linking to any RFCs or The Code Style guide/[dev docs](https://docs.geomoose.org/3.x/developer.html) in here?

Should there be comments about contributors attest that they have the right to contribute their changes under LICENSE?  I think this is partially covered in the [AUTHORS](https://github.com/geomoose/gm3/blob/main/AUTHORS) file and [RFC-4](https://www.geomoose.org/rfc/rfc-4.html), but is a bit different than this doc states as "All are welcome"... it's more of "All are welcome who follow the process".   But we are a little unclear here, because I think from the wording of the existing docs, they mostly apply to people who can merge into main, not necessarily anyone who can contribute a PR.

How do we assess if AI assisted contributions can properly attest they have the right to contribute the changes under the GeoMoose LICENSE?

We should also probably mention SECURITY.md.  "Bug reports welcome as patches" is in tension with please follow SECURITY.md for responsible disclosure for security related bugs, so I think we should at least mention the SECURITY.md route exists so people don't just stop at reading this file if this is really a security report.

jmckenna:
I added a link to the PSC page.

jmckenna (approved):

elil:
@theduckylittle At the PSC meeting, folks seemed to think this required a vote for setting project policy.  There's email list discussion as well.  I made a motion but then retracted it due to subsequent changes.

PR #1021, opened 2026-08-18
fix: Only print the scale label when the map rendered at that scale

theduckylittle:
@theduckylittle messed up the merges of #999 and #1000 so they landed on main. This PR is an attempt to address the follow up comments for that.

## 1. A printed scale label that isn't true

@chughes-lincoln, [#999 (Aug 13)](https://github.com/geomoose/gm3/pull/999#issuecomment-5285526341):

> Something that might be worth mentioning in the docs - if the admin has configured zoom restrictions, they need to make sure that they adjust the print to scale options so that they're compatible with their zoom restrictions.
>
> i.e. if I've set a maxZoom of 20, and I update to the geomoose version that contains this PR without changing any of the default settings, users will be able to print to 10ft/in. It will still print with a label saying 1 inch = 10 feet, but in this case the result is closer to 1 inch = 75 feet.

I've added a check to available scales that prevents the over-zoom! The user won't be able to get to options they cannot "use." This can create some odd hidden states because the user may not see something the admin configures.

The "scale label" will also only display if it actually matches the map. This does not solve the misleading scale lines issue...

## 2. Misleading scale lines

@chughes-lincoln, same comment:

> I also heard feedback about wanting to be able to export a pdf at a particular scale, while also omitting the 1 inch = 10ft label. The reason being to avoid any complaints/liability for when people take a pdf exported at a particular scale, and then print it at a different one (for example I printed something at Letter - Portrait, but the printer settings defaulted to 11x17 Portrait with Fit to printable area checked - so the 10ft scalebar was still correct because it didn't change the aspect ratio, but it was no longer 10ft/inch).

@klassenjs, [#999 (Aug 17)](https://github.com/geomoose/gm3/pull/999#issuecomment-5318685424):

> Something that says something like "the scalebar should measure exactly 1 inch when printed at the intended scale" can help with this. Using a 2D reference feature (a square somewhere or maybe even the map frame) could also show issues with aspect ratio. But I could see where people might want to just turn that off too (or say not to any particular scale).

I've deferred this to being a user choice. There is now a scale line flag that will either show:

- None. Don't display a thing
- Line and Distance. Displays `|___ 100 ft ___|` (excuse my most excellent ascii art)
- Line, Distance, and Label which will display the above plus any relevant given ratio.

Users can still do awful things but at least they have some safeguards from doing so. This also prevents us from removing a user choice that may be entirely valid. I'm thinking about maps in which the user would want, say, the 1:50 scale to be shown as it is a requirement of submitting the map.

### DIFF
--- a/docs/howto/configure-print-scales.rst
+++ b/docs/howto/configure-print-scales.rst
@@ -82,3 +82,48 @@ Scale                     ``scale``    Typical use
 
     Scales assume a basemap in meters (web mercator) and are true at the
     map center; like any web-mercator map, scale distorts toward the poles.
+
+Zoom restrictions and the offered scales
+----------------------------------------
+
+Printing at a fixed scale re-renders the map at the resolution that scale
+requires. The print map is rendered by the same map component as the main
+map, so it inherits any ``map.view`` zoom restrictions -- and a scale
+finer than those restrictions permit **cannot be honored**.
+
+::
+
+    var app = new gm3.Application({
+        map: {
+            view: {
+                maxZoom: 20
+            }
+        }
+    });
+
+With ``maxZoom: 20``, a request for ``1 inch = 10 feet`` cannot be met:
+OpenLayers clamps the render at the zoom-20 limit, and the map comes out
+several times coarser than asked for.
+
+The same applies at the other end. ``view.extent`` caps how far out the
+map can zoom -- the extent has to fit the printed map's pixel size -- so
+a very coarse scale on a tightly bounded map is equally unreachable.
+
+**Unreachable scales are not offered.** GeoMoose tests each configured
+scale against the view's limits and leaves out the ones it cannot honor,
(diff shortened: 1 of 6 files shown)


chughes-lincoln:
I'm getting a minified react error in the console (on page load, didn't need to print anything), which points me to this webpage - https://react.dev/errors/130?invariant=130&args%5B%5D=undefined&args%5B%5D= (Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined.) This was both without making any changes to the demo, and with adding a default printMode in app.js

Everything else I've tested so far works as expected.

theduckylittle:
@chughes-lincoln , I'm not getting that error?  Can you attach your app.js?

chughes-lincoln:
Upon further review that was an error on my end. I failed to overwrite the app.js from the feature report PR. So it was throwing an error because the ReportModal didn't exist.

Issue #1022, opened 2026-08-19
Print hangs with OpenStreetMap display errors (and other display errors?)


chughes-lincoln:
On the demo if you zoom in past ~50ft/in, OpenStreetMap stops loading and you get a bunch of CORS errors in the console.

I can consistently get the print modal to hang when changing resolutions at that scale. 

Steps to reproduce
- Open the demo
- Click Print
- Switch the resolution to 10 ft / in - this will give you CORS errors in the console, but printing at this stage still works
- Switch to a different resolution at 50 ft / in or smaller
- Go make a coffee
- See that the print modal is still spinning - the user can still change the layout and resolution, which depending on the choice stops things from hanging

<img width="938" height="777" alt="Image" src="https://github.com/user-attachments/assets/d52eab95-596c-4e6a-a4dd-82b27fa08a42" />

I'm less consistently running into the same issue with other basemap datasets. It may be more related to changing the resolution when no data exists in the preview. I switched to No Background and turned off the parcels layer. Once I go to print and change the resolution, pretty much every other ft/in option hangs. It doesn't occur with the scale to fit options. I don't see any noteworthy console or Apache errors (other than the CORS errors for OSM)

PR #1029, opened 2026-09-03
fix: Default to blank image on tile error load

theduckylittle:
This addresses issues when loading high resolution tiles on layers which do not support it.

When printing, a tile layer can be fetched past its maximum resolution or create error states when the layer is not properly configured.

refs: #1022

### DIFF
--- a/src/gm3/components/map/layers/xyz.js
+++ b/src/gm3/components/map/layers/xyz.js
@@ -26,12 +26,26 @@
  *
  */
 
-import XYZSource from "ol/source/XYZ";
 import TileLayer from "ol/layer/Tile";
+import TileImageSource from "ol/source/TileImage";
+
+// this is a blank base64 encoded png, to avoid cors issues (especailly while printing)
+//  tiled layers default to a blank tile on error
+const BLANK_PNG =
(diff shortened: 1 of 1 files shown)


chughes-lincoln:
I'm still seeing the same behavior.



https://github.com/user-attachments/assets/77b6595e-097b-4d08-bac8-3d3a4953fb38

chughes-lincoln (commented):
Seeing same behavior (CORS error at high resolution in console, hangs when switching from 10ft/in to 20ft/in)

theduckylittle:
There is no way to prevent the CORS errors from happening in the console unless the min/max resolution is set for the layer. This is something we could do in the demo but don't currently.

I'll re-check on the printing issue it may be a different layer causing problems but in my testing, things were working.... :thinking:

chughes-lincoln:
I'm pretty sure the print hanging happens when there's no data in the print preview. I'm guessing that it was working for you when you were testing because the parcels were still on, and by chance the parcels were at the print location.

On this PR if I manually zoom in to one of the parcel boundaries (so that I can see a green line in the print preview for 10/20/40/50 ft/in), nothing hangs. If I pan to a spot where there are no parcels (or turn them off), it hangs.


https://github.com/user-attachments/assets/41569cd5-c46d-441f-bcf2-d9494e26dbfd

PR #1032, opened 2026-09-03
fix: Better "start over" behavior

theduckylittle:
Fixes the bug identified with #993 (`mapbook=` being cleared), but provides a more complete solution based on the known parameters used by services on startup.


chughes-lincoln:
Works for me as well.

